PR #6, opened 2025-10-02
Add Features to PSLF Plug-in

sjenkinsnm:
Working on getting a tool that is capable of completing a college level power systems study. To this end, added the ability to change the model and run contingency analysis to the tool.


muhi-zatar (approved):
Looks good!
If you can or see necessary to implement validators on the input, please do.
But It can be merged in its current form.

PR #9, opened 2025-10-18
Add MCP entry point for NARP reliability tools

839651726:
# Summary
- Introduce NARP/narp_mcp.py. It exposes one FastMCP server and registers the tool functions with power_mcp_tool, following the layout used by the other integrations.

# Why
- This change lets PowerMCP users call the NARP reliability analysis server directly from the repo. The workflow now matches the other MCP tools, and the README gives concrete prompts for MCP clients.

# Testing
- Manual run: python -m NARP.narp_mcp to confirm the FastMCP server starts and registers the tools.
- Use the README prompts in an MCP-enabled client to call validate_input, run_simulation_sync, etc., and check the responses.


839651726:
Thanks for the review. I updated the branch to address both comments.

## Case files cleanup
I trimmed the bundled NARP data to a minimal example set and removed large test/integration datasets from this PR, so the integration stays clean and lightweight.

Now the branch keeps one runnable sample case at `NARP/examples/case1` with only the required input artifacts:
- `ZZMC.csv`
- `ZZUD.csv`
- `ZZLD.csv`
- `ZZTD.csv`
- `ZZTC.csv`
- `LEEI`

## NARP introduction for non-reliability users
I added a **“What is NARP?”** section in `NARP/README.md` with a concise overview of:
- What NARP does (resource adequacy / reliability assessment)
- Key indices (LOLE, HLOLE, EUE)
- Typical planning and maintenance-study use cases

I also updated usage examples to reference the minimal bundled case and aligned the MCP server README examples accordingly.

qian-harvard:
@839651726 Can you update your PR based on the new main branch?

qian-harvard:
Thank you for this, @839651726, and I'm sorry it sat as long as it did — that's on us, not you.

Two things I want to name first. You turned my March review around quickly and did both things I asked: the minimal `examples/case1` set and the plain-language LOLE/HLOLE/EUE introduction in `NARP/README.md`. That introduction is genuinely good and I'd like to reuse it. And NARP is still the only resource-adequacy capability anyone has proposed for this repo — nothing on main or in any open PR does sequential Monte Carlo LOLE/HLOLE/EUE.

I'm closing the PR, but not the idea. #74 carries your work forward with exact revisions and git references, so whoever picks it up starts from what you built rather than a blank page.

**1. The branch predates the current repo layout, and most of that is our doing.** The merge base is `ba26bad` (2025-09-17). We landed `699dff3` "PowerMCP 1.0" on 2026-03-14 — the same day as the first review — which deleted the root `common/` package, so `narp_mcp.py:14`'s `from common.utils import ...` breaks the moment this merges. The packaging layer `f994a15` (`pyproject.toml`, `powermcp/registry.py`) landed 2026-06-09, six weeks after your last push, so there was no registry to register NARP in while you were working. Neither counts against you.

**2. It targets the MCP 1 API, which is gone.** `narp_mcp.py:15` and `common/utils.py:4` import `mcp.server.fastmcp`. Main pins `mcp>=2,<3` and all 29 server modules now import `mcp.server.mcpserver`.

**3. The job runner never actually runs NARP.** This is the decisive one, and I checked it by running it rather than reading it. `mcp_server/worker.py:67-69` launches `reliabilityassessment/monte_carlo/narpMain.py` as a script, but that file defines `narpMain(TEST_DIR)` with no `if __name__ == "__main__":` guard. Executed exactly as the worker does it:

```
exit code: 0
stdout bytes: 0
files written to job dir: 0
```

`worker.py:77` raises only on a nonzero exit, so `:82-90` then marks the job `completed` and returns an `output_path` that was never created. The `-m` fallback at `:71` has the same problem. Worth saying that upstream has the same gap — its own integration test comments the call out with `run "narpMain(TEST_DIR)" externally` — so this is an integration oversight, not carelessness.

**4. The bundled example can't complete even when `narpMain` is called correctly.** `readInputB.py:541-543` unconditionally reads `ZZFC.csv`, `ZZOD.csv` and `ZZDD.csv` (at `:415`, `:460`, `:497`), none of which `case1` contains. `_REQUIRED_INPUT_FILES` at `narp_mcp.py:29-36` doesn't list them, so `validate_input` returns `{"valid": true}` for a case that crashes on the next call. Each is a small fix, but together they mean the connector has never produced a result end to end.

**5. A licensing question — and I mean it as a question.** `NARP/reliabilityassessment/` is Breakthrough-Energy/reliability-assessment, and the attribution is clearly there: `docs/index.rst:3` names Dr. Chanan Singh and Dr. Yongli Zhu, `setup.py:9` links upstream, `docs/installation.rst:9` and `docs/manual.rst:4` do too, and `README.md` lists Dr. Zhu as a contact. Nothing was removed and I'm not suggesting otherwise.

What I can't find anywhere under `NARP/` is upstream's `MIT-LICENSE.txt`. That matters more here than in a plain fork, because PowerMCP publishes a wheel and `pyproject.toml`'s `force-include` maps each server directory into it verbatim — so adding NARP as-is would redistribute upstream's code on PyPI without its license text. Had you had a chance to check with upstream how they'd want that handled?

**What would land.** Don't vendor. For the four engines whose upstream isn't pip-installable, this repo uses a thin connector plus a minimal or empty extra with manual install documented in the connector README (`psse = []`, `pslf`, `powerfactory`, `plexosdb`). A direct git dependency isn't open to us either — see the `plexosdb` comment in `pyproject.toml`: PyPI rejects packages declaring one, which would break `pip install powermcp` itself. That shape takes this from 95 files and 13,772 lines down to roughly the size of the ANDES connector, removes the licensing question entirely, and is reviewable in an afternoon.

**What I'm keeping.** Your `parse_output()` TABLE 13 mapping is correct — it picks the MAGN columns and skips the interleaved PCT SD ones, which is real domain knowledge and expensive to rediscover. The submit/run-sync split is a sound shape for wrapping a long Monte Carlo run. Both are recorded in #74 with git revisions, along with the reference NARP report from your branch history — the only TABLE 12/13 ground truth in this project.

Thank you again. I'd like to see the thin version.

PR #25, opened 2026-05-01
Added run project tool call with status checks

LeoWiniecki-PowerSystems:
Added the ability to run a simulation and wait for the result.

### DIFF
--- a/PSCAD/pscad_mcp/core/connection_manager.py
+++ b/PSCAD/pscad_mcp/core/connection_manager.py
@@ -1,5 +1,6 @@
 import psutil
 import logging
+import time
 from typing import Optional
 import mhi.pscad
 from pscad_mcp.core.executor import robust_executor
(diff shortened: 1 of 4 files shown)


qian-harvard:
@LeoWiniecki-PowerSystems Can you update your PR based on the new main branch?

qian-harvard:
Thank you for this, @LeoWiniecki-PowerSystems, and apologies for the slow verdict.

I'm closing the PR, but one part of it is better than what main has today and I've filed it to be taken: `set_component_parameters_safe` and `values_equivalent`. Main's `set_component_parameters` (`PSCAD/pscad_mcp/tools/project_tools.py:130-136`) sets a value and returns `"Parameters updated for component {id}."` with **no read-back at all** — so an agent is told "updated" even when PSCAD silently clamps or rejects the value. Your version sets, reads back, compares unit-aware (`"100 [ohm]"` vs `100`), and rolls back to a pre-change snapshot on mismatch, with three passing tests. That's a real bug fix and it applies cleanly to main. It's specced in #76 with a `Co-authored-by:` note for you.

Reasons for closing:

**1. Staleness.** I asked for a rebase on 2026-06-10 and haven't heard back. The merge base is `d7db0df` (2026-04-25), 100 commits behind, with conflicts in three files. That's the plainest reason and I'd rather lead with it than dress it up.

**2. Main reimplemented the underlying mechanism.** #48 (`0a3919d`) moved PSCAD to non-blocking `run_project` + `get_run_status` polling, plus `build_project`/`get_build_messages`. What this PR still adds over main is ergonomic — one tool call instead of agent-driven polling — plus MCP progress streaming, which main genuinely lacks (`report_progress` has zero matches repo-wide). That delta is real, which is why it's now #76 rather than discarded. But it isn't worth rebasing 475 lines onto a different set of primitives.

**3. `run_project_and_wait` would fail on a real simulation.** `project_tools.py:125` calls `run_safe(project.run)`, and `Project.run()` blocks for the whole build-and-run, so under the 30 s watchdog it times out while holding the single PSCAD worker. I'm taking the blocking behaviour from `0a3919d`'s own end-to-end verification rather than from the library — I don't have PSCAD here.

**4. A failed build is reported as a successful simulation.** This one is visible in your code without PSCAD, and it's the one I'd most want fixed in any revival: the poll loop exits when `run_status` returns `None`, and the function then returns `{"final_status": "completed"}` unconditionally. A build that fails to compile never reaches `running`, drops straight back to idle, and the agent is told the simulation finished. The tests can't catch it because the fake's `run_status` script never models that path.

**A fairness point that belongs in this comment.** You pushed on 2026-05-01. The commit that discovered `Project.run()` blocks and that the watchdog needed a `_timeout` override is `0a3919d`, dated 2026-06-23 — seven weeks later. Worse, this repo's own `PSCAD/COMPARISON_GUIDE.md:12-18` showed *exactly* the pattern you implemented, and `:34` advertised the watchdog over `project.run()` as a feature — and **it still says that on main today**, byte-identical to what you read. You coded against the documentation we published. I've filed #75 to fix it. These blockers are the cost of a stale branch and of our own wrong docs, not of carelessness on your part.

#76 has the design on main's primitives — `project.start()`, `_timeout=0` for the wait, a `saw_running` flag, `get_build_messages` to tell a failed build from a finished run — where both blockers disappear by construction. Your poll-backoff schedule is carried over and credited. If you want that one, it's yours.

Contributors with a real PSCAD install are the scarcest thing this subtree has, and you're one of them. I hope you'll come back.

PR #34, opened 2026-06-10
fix(powerio): out-of-service handling in PyPSA bridge + error-shape hardening

qian-harvard:
Follow-up fixes from an adversarial review of the powerio integration.

- PyPSA bridge: drop out-of-service branches/generators before import_from_pypower_ppc (which ignores ppc status) so switched-out elements no longer import as active; corrected the rating-0 warning; clearer write-vs-parse error. pandapower's from_ppc honors status, so it is unchanged (documented by a new test).
- powerio server: normalize convert_case OSError to ValueError; map wrong-schema JSON to the documented parse error; fail-fast namespace-shadow guard; explicit laplacian dispatch.
- 6 new regression tests; full suite 91 passed.

Targets 0.1.2.

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PyPSA/pypsa_mcp.py
+++ b/PyPSA/pypsa_mcp.py
@@ -667,14 +667,41 @@ def _import_case_to_netcdf(case, output_path: str, overwrite_zero_s_nom: Optiona
         )
     if (ppc["bus"][:, 9] == 0).any():
         warnings.append("buses with base_kv 0 are assigned v_nom 1 by PyPSA")
-    in_service = ppc["branch"][:, 10] == 1.0
-    if overwrite_zero_s_nom is None and (ppc["branch"][in_service, 5] == 0).any():
+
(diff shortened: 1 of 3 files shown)


samtalki:
Nice catch — `import_from_pypower_ppc` really does ignore the status columns, and the gencost realignment after dropping generators is a detail I'd have missed. Thanks!

Two coordination notes:
- The py3.10 failure here is preexisting on main (a doctor test asserts surge is "missing", but on 3.10 the version gate answers first). The fix is on #35, so whichever PR lands first, the other goes green on rebase.
- #35 replaces `powerio/powerio_mcp.py` with a copy of the canonical upstream server (powerio 0.1.0: `convert_str`, no temp staging), so our PRs overlap on that file and the test file. Happy to rebase #35 over this once it merges and carry your hardening along — the shadow-import guard fits the copy here, and I'll send the shareable parts upstream so the two files stay in sync.

PR #28, opened 2026-06-10
feat: integrate powerio as the cross-server case exchange substrate

samtalki:
## What this adds

powerio becomes the format conversion substrate connecting PowerMCP's servers. The integration has three pieces:

**The powerio server** (`powerio/powerio_mcp.py`, registry id `powerio`): a standalone FastMCP script with seven tools. `convert_case` and `case_summary` convert and summarize any powerio readable case (MATPOWER `.m`, PSS/E `.raw` v33, PowerWorld `.aux`, PowerModels JSON, egret JSON). `parse_case`, `normalize_case`, and `case_to_json` emit the powerio JSON transport (`Network.to_json()`). `compute_matrix` returns B', B'', Re/Im(Y_bus), adjacency, PTDF, LODF, or the weighted Laplacian in COO form as plain Python lists; `dense_view` returns the dense table view.

**Bridge tools** in the existing servers (added only; no existing signature changed):
- pandapower: `load_network_from_any`, `load_network_from_json`, `export_network_to_format`. Loading goes powerio → PYPOWER ppc dict → `pandapower.converter.pypower.from_ppc` (`from_mpc` reads only binary `.mat`, so the ppc route is the supported path); export goes `to_ppc` → MATPOWER text → powerio.
- egret: `load_model_from_any`, `load_model_from_json`. powerio converts to egret JSON, the parsed dict constructs a `ModelData` to validate, and the JSON is staged to a temp file whose path feeds the existing solver tools.

**Packaging**: a `powerio` extra (`powerio[mcp,matrix]`), membership in the `opensource` group, the wheel force-include (`powermcp/_servers/powerio`), and the sdist include. `powermcp list` and `powermcp doctor` pick the tool up from the registry entry.

## The exchange pattern

Parse a case once, pass the JSON string between tool calls, load it anywhere:

```
parse_case(path="case9.raw")                 # powerio server → {"json": ..., "summary": ...}
load_network_from_json(network_json=...)     # pandapower server ingests the transport
load_model_from_json(network_json=...)       # egret server stages it as a solvable case file
compute_matrix(kind="ptdf", json=...)        # powerio server builds matrices from it
```

Downstream tools reconstruct with `powerio.from_json(...)`; no re-parse, no shared filesystem needed between servers.

## Graceful degradation

powerio is an optional extra. Every bridge tool catches the ImportError and returns `{"status": "error", "message": "powerio not installed: pip install 'powerio[mcp,matrix]'"}`; the pandapower and egret servers work unchanged without it.

[...]

PR #35, opened 2026-06-10
feat: sync the powerio server with upstream 0.1.1

samtalki:
Follow-up to #28, against the released powerio 0.1.1 ([release notes](https://github.com/eigenergy/powerio/releases/tag/v0.1.1)).

- **Inline conversion is now in memory**: powerio 0.1.0 ships `convert_str`, so `convert_case` no longer stages inline content to temp files; the `_EXT`/`_stage`/`_unlink_quietly` machinery is deleted and a regression test asserts the tool never touches `tempfile`.
- **The server file is now a faithful copy of the canonical upstream** (`powerio.mcp.server`, declared canonical in eigenergy/powerio#90): same eight tools, plus the docstring improvements from its review (accurate `json=` acceptance lists, "reference bus index" in `dense_view`, tightened module docstring). The standalone header points back at upstream so future syncs have a direction.
- **The `powerio` extra pins `>=0.1.0`** (older versions lack `convert_str`).
- **Includes the 0.1.1 error-shape hardening** (eigenergy/powerio#93, the shareable parts of #34): an unreadable input maps to the documented ValueError shape; the upstream suite's unreadable-file and wrong-schema tests are ported.

Suite: 85 passed, 1 skipped (the Windows-only detect test, per #29's skipif). The server still runs standalone (`python powerio/powerio_mcp.py`) and `powermcp list`/`doctor` are unchanged.

### DIFF
--- a/ANDES/andes_mcp.py
+++ b/ANDES/andes_mcp.py
@@ -8,7 +8,7 @@ import json
 from pathlib import Path
 from contextlib import redirect_stdout, redirect_stderr
 from mcp.server.fastmcp import FastMCP
-from typing import Dict, Any
+from typing import Dict, Any, Optional
 
 # Storage directory resolved lazily (no filesystem writes at import time)
 def _andes_runs_dir():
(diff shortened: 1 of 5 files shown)


samtalki:
This PR is now rebased on main and includes all of Qian's fixes from #34 (PyPSA OOS drop, shadow-import guard, error normalization, laplacian elif). 91 tests pass on Python 3.12.

One thing worth flagging: powermcp 0.1.1 shipped with a silent topology bug in the PyPSA bridge. `import_from_pypower_ppc` ignores the ppc status columns, so out-of-service branches and generators were imported as fully active — wrong topology with no warning. That's fixed here (and would have been fixed by #34 too). Apologies for missing this before the 0.1.1 release.

If you agree it warrants a patch, a **0.1.2** release after this merges would be a clean way to get the fix to users. Happy to do whatever's most useful — just mention it and I'll handle the prep.

samtalki:
@qian-harvard — carried your OOS fix and error hardening from #34 into this branch (see `73ee997`). The ANDES bridge added in the latest commit follows the same pattern you established for PyPSA. Would love your eyes on the bridge section in `ANDES/andes_mcp.py` if you get a chance!

PR #61, opened 2026-08-28
Update PSCAD MCP: project_tools.py and simset_tools.py

Xiaoyang-Wang-TAMU:
**PSCAD** MCP Tool modification:
find_components is fixed with correct parameters.
Passed `definition` and `name` incorrectly as keyword arguments to `Project.find_all()`; the installed API expects positional arguments

simset_tools.py is fixed by replacing all project.simset with workspace.simset


PSCAD/pscad_mcp/tools/project_tools.py
```diff
@@ -116,8 +116,51 @@ async def find_components(
...
+    if name:
+        criteria.append(name)
+
+    if not criteria:
+        raise ValueError(
+            "At least one of 'definition' or 'name' must be provided."
+        )
+
+    components = await robust_executor.run_safe(
+        project.find_all,
```
qian-harvard, line 136: Passing the two criteria positionally means the library's disambiguation heuristic decides what they mean, and it can silently flip them. From `mhi/pscad/project.py:1404-1417`:

```python
if defn:
    if defn in _BUILTIN_COMPONENTS:       # 'Bus', 'TLine', 'Cable', 'GraphFrame',
        pass                              # 'Sticky', 'Wire', 'Divider', ...
    elif ':' in defn:
        namespace, defn = defn.split(':', 1)
    elif not named:                       # <-- lone bare positional
        named = defn                      #     becomes a NAME search
        defn = None
```

Two consequences:

- `find_components(name="Bus")` → the lone positional `"Bus"` is a builtin, so it becomes a **definition** search and returns every bus bar in the project instead of the component named `Bus`.
- `find_components(definition="source3")` → not a builtin, no colon, no `named` → it collapses into a **name** search and returns `[]` instead of every `master:source3` instance.

Neither raises, so the caller just gets wrong data.

The name-only case has a clean fix — pass `None` in the definition slot so the `if defn:` block is skipped entirely:

```python
if definition and name:
    components = project.find_all(definition, name)   # this ordering is correct
elif definition:
    components = project.find_all(definition)
else:
    components = project.find_all(None, name)         # pins `name` to the name slot
```

The definition-only case can't be fixed the same way — the API has no way to express "this bare string is a definition" (passing `find_all(definition, None)` collapses identically, since `named` is falsy). So `definition` genuinely has to be namespace-qualified (`master:source3`) or one of the builtins. Worth stating that in the docstring, since the docstring becomes the MCP tool description the client reads:

```python
"""Find components matching criteria in a project.

``definition`` must be namespace-qualified (e.g. ``master:source3``) or a
built-in name ("Bus", "TLine", "Cable", "GraphFrame", "Sticky", ...). A bare
definition name is interpreted by the PSCAD API as a component name.
"""
```

If you want to be stricter, rejecting an unqualified non-builtin `definition` with a clear error beats silently degrading it to a name search.

PSCAD/pscad_mcp/tools/project_tools.py
```diff
@@ -116,8 +116,51 @@ async def find_components(
...
+            or parameters.get("name")
+            or parameters.get("NAME")
+            or ""
+        )
+
+        results.append(
+            {
+                "id": component.iid,
+                "name": component_name,
+                "definition": component.defn_name,
```
qian-harvard, line 158: This one looks like a plain attribute read but it isn't. From `mhi/pscad/component.py:231-236`:

```python
@rmi_property
def _defn_name(self): ...

@cached_property
def defn_name(self) -> str:
```

So the first access per component is a synchronous RMI round-trip — and it's happening on the asyncio event-loop thread, outside `run_safe`. That bypasses all three things the executor exists to provide: the single COM-initialized worker, the lock, and the watchdog. If PSCAD is frozen or showing a modal dialog here, the event loop itself blocks and no watchdog can fire, so the whole server hangs for every client rather than one call timing out.

There's also a transport hazard: `mhi/common/remote.py:1007-1012` is `self._write(msg)` then `self._response.get()` on a single shared queue with no correlation IDs and no send lock, so a `defn_name` read racing an in-flight worker call can consume the wrong response.

(`component.iid` on line 156 is fine — it's a local `self._identity['iid']` read, no RMI.)

Easiest fix is the batching one in my other comment, which puts both the `parameters()` and `defn_name` reads on the worker thread.

PSCAD/pscad_mcp/tools/project_tools.py
```diff
@@ -116,8 +116,51 @@ async def find_components(
...
+            }
+        )
```
qian-harvard, line 160: Two things here.

**Missing `None` guard.** `component.parameters()` can return `None` — `FormCodec.decode` passes it straight through (`mhi/pscad/form.py:435-436`), and this file already defends against it in `get_component_parameters`:

```python
return params if params else {}   # project_tools.py:171
```

Without that guard, one component returning `None` raises `AttributeError: 'NoneType' object has no attribute 'get'` and kills the whole listing, discarding every result already collected.

**One dispatch per component.** Each iteration is a separate `run_safe` — a thread hop, a lock acquire/release, and its own 30s watchdog — plus a `defn_name` RMI. A search matching a few hundred components (easy with `definition="Bus"` on a real case) becomes a few hundred sequential round-trips, and total latency is unbounded even though no single call trips its timeout.

Both go away if the loop runs inside one `run_safe` closure on the worker thread:

```python
def collect():
    results = []
    for component in components:
        parameters = component.parameters() or {}
        results.append({
            "id": component.iid,
            "name": (parameters.get("Name")
                     or parameters.get("name")
                     or parameters.get("NAME")
                     or ""),
            "definition": component.defn_name,
        })
    return results

# Covers N round-trips, so give it the same generous budget as load_projects.
return await robust_executor.run_safe(collect, _timeout=120)
```

One aside: the `Name`/`name`/`NAME` chain is correct, not defensive guesswork — `find_all`'s own docstring (`project.py:1370-1371`) defines a component's name as "a parameter called `name`, `Name`, or `NAME`". Maybe add a one-line comment saying so, since it reads as speculative otherwise.

PSCAD/pscad_mcp/tools/project_tools.py
```diff
@@ -116,8 +116,51 @@ async def find_components(
     """Find components matching criteria in a project."""
     pscad = pscad_manager.pscad
     project = await robust_executor.run_safe(pscad.project, project_name)
-    components = await robust_executor.run_safe(project.find_all, definition=definition, name=name)
-    return [{"id": c.id, "name": c.name, "definition": c.definition} for c in components]
+    # components = await robust_executor.run_safe(project.find_all, definition=definition, name=name)
+    # return [{"id": c.id, "name": c.name, "definition": c.definition} for c in components]
+    ####### Code is modified
```
qian-harvard, line 121: Could you drop the commented-out old implementation and the `####### Code is modified` / `########` banners? Git history already has the old version, and the commented block references attributes that don't exist on the real API (`c.id`, `c.name`), so it's actively misleading to anyone who tries to restore it later.

The trailing `     ########` on line 163 also sits at a stray 5-space indent after `return results` — harmless as a comment, but a future edit aligned to it would be an `IndentationError`.

PSCAD/pscad_mcp/tools/project_tools.py
```diff
@@ -116,8 +116,51 @@ async def find_components(
...
+    if definition:
+        criteria.append(definition)
+
+    if name:
+        criteria.append(name)
+
+    if not criteria:
+        raise ValueError(
+            "At least one of 'definition' or 'name' must be provided."
+        )
```
qian-harvard, line 133: Minor: this is equivalent and matches the single-line `run_safe` style used elsewhere in the file (e.g. line 169):

```python
criteria = [c for c in (definition, name) if c]
```

Also worth knowing: the no-criteria `ValueError` you added mirrors the library's own behavior — `find_all` raises `ValueError("No search criteria given")` at `project.py:1398-1399` — so you're not removing a "list all components" mode. Good call.

PSCAD/pscad_mcp/tools/simset_tools.py
```diff
@@ -4,29 +4,53 @@
...
+
 
-async def run_simulation_set(project_name: str, sim_set_name: str) -> str:
-    """Run a specific simulation set (batch of tasks)."""
+async def run_simulation_set(
+    project_name: str,
+    sim_set_name: str,
+) -> str:
+    """Run a simulation set and wait for all of its tasks to finish."""
     pscad = pscad_manager.pscad
```
qian-harvard, line 22: `run_project` refuses to start when the license isn't available (`project_tools.py:37-38`):

```python
if not await robust_executor.run_safe(pscad.licensed):
    return {"started": False, "error": "PSCAD is not licensed."}
```

Worth adding the same guard here. Without it, an unlicensed PSCAD can stall or pop a licensing dialog inside a call that has its watchdog explicitly disabled — so the single worker blocks indefinitely on a batch that was never going to run. One call turns that into an immediate, clear error.

qian-harvard (commented):
Thanks for digging into these — I checked both fixes against the actual `mhi-pscad` 3.1.2 source (pulled the wheel from PyPI), and you're right on both counts. `Project` has no `simulation_set`/`simulation_sets` methods at all; they only exist on `PSCAD` (`pscad.py:1497`, `:1549`), so the old simset tools would have raised `AttributeError` against a real PSCAD. And `find_all` really does take positional criteria — the old `definition=`/`name=` kwargs were falling into `**params` and being matched as literal parameter names, plus `c.id`/`c.name` don't exist on components (they're `iid`/`defn_name`). So all three of these tools were broken on `main` and only looked healthy because the tests mock them.

A few things to sort out before this can merge.

## Blocking

1. **Two tests fail on this branch** (`main` is 17/17, this is 15/17) — see below.
2. **`find_components` can silently return the wrong result set**, because of how `find_all` reinterprets a lone positional argument. (inline)
3. **`component.defn_name` is a remote call made outside `robust_executor`.** (inline)

## Worth discussing

4. **`run_simulation_set` holds the single executor worker for the whole batch with the watchdog off**, which starves every other tool. (inline)
5. **`project_name` is still a required schema argument** on tools that immediately `del` it. (inline)

---

### The two failing tests — `PSCAD/tests/test_enhanced_tools.py`

These are mock-based, so they run fine without PSCAD installed:

```
FAILED test_enhanced_tools.py::TestEnhancedPSCADTools::test_list_simulation_sets
FAILED test_enhanced_tools.py::TestEnhancedPSCADTools::test_run_simulation_set
```

- **`test_list_simulation_sets`** — line 83 mocks `self.mock_project.simulation_sets`, but the tool now calls `pscad.simulation_sets`, so the mock is never reached: `AssertionError: 'Batch1' not found in <MagicMock name='pscad_manager.pscad.simulation_sets()'>`. It also needs to return plain strings now, since `PSCAD.simulation_sets()` is typed `List[str]` (`pscad.py:1497`) rather than objects with `.name`.

- **`test_run_simulation_set`** — line 94 asserts `"started"` but the message now says "completed". Heads up that there's a *second*, independent break behind that one: `setUp` line 27 wires the fixture as `self.mock_project.simulation_set.return_value = self.mock_sim_set`, so with the tool now resolving via `pscad.simulation_set`, `.run()` lands on an auto-created child of `mock_pscad` and line 95's `self.mock_sim_set.run.assert_called_once()` fails with "Called 0 times". Fixing only the assertion string will leave it red.

Roughly:

```python
# setUp
self.mock_pscad.simulation_set.return_value = self.mock_sim_set

# test_list_simulation_sets
self.mock_pscad.simulation_sets.return_value = ["Batch1"]
```

Also worth adding: no test anywhere mocks `find_all` returning actual components (`test_tools.py:88` returns `[]`), so the entire new `iid`/`defn_name`/`parameters()` path is uncovered.

### Stale docs

- `PSCAD/README.md:113` still says `list_simulation_sets` discovers batch runs "in a project", and `PSCAD/EXAMPLES.md:27` says "List all simulation sets in the active project" — both now workspace-wide. Since EXAMPLES is a prompt guide, an LLM client reading it will present workspace-wide results as project-specific.
- `PSCAD/COMPARISON_GUIDE.md:34` now overclaims: "Watchdog Protection: If `project.run()` takes too long to respond, the MCP Executor triggers a timeout instead of hanging the AI" — no longer true for `run_simulation_set`, the one tool that disables the watchdog.

### `PSCAD/pyproject.toml:23` — unpinned `mhi-pscad`

Not introduced by this PR, but this PR is the first code to depend on it: `defn_name` changed return type in **3.0.2** (`component.py:236-249`: "Return type changed from `Union[str, Tuple[str, str]]` to `str`"), so on an older install `find_components` emits `"definition": ["master", "source3"]` instead of `"master:source3"`. Since the extra is just `windows = ["mhi-pscad", ...]` with no bound, a `>=3.0.2` lower bound would make this safe. Everything else the PR uses is 2.0+.

---

Happy to pair on the sim-set piece if you'd rather split it into its own PR — that one has a real design question in it.

PSCAD/pscad_mcp/tools/simset_tools.py
```diff
@@ -4,29 +4,53 @@
...
-async def add_task_to_set(project_name: str, sim_set_name: str, task_project_name: str) -> str:
-    """Add a project task to an existing simulation set."""
+    sim_set = await robust_executor.run_safe(
+        pscad.simulation_set,
+        sim_set_name,
+    )
+
+    # A simulation set may run much longer than the default 30-second
+    # watchdog, so disable the watchdog for this blocking API call.
+    await robust_executor.run_safe(sim_set.run, _timeout=0)
```
qian-harvard, line 30: This is the one I'd most like to talk through. `SimulationSet.run()` blocks until every task finishes (`simset.py:362` wraps a blocking `_run` RMI), and `RobustExecutor` has exactly one worker plus a lock (`executor.py:37, 57`). So for the entire duration of a batch — potentially hours — every other tool call queues behind this and then dies at its own 30-second watchdog with:

> PSCAD timed out during X. It might be frozen or showing a dialog.

...which is a confusing thing to tell a user whose simulation is running perfectly fine. `get_run_status`, `list_projects`, even `repair_connection` are all affected, and there's no cancel path: cancelling the coroutine can't stop a thread already running in the executor.

This is the exact failure mode `run_project` was written to avoid — its docstring says it uses the non-blocking `start()` specifically to avoid "holding the single PSCAD worker for the entire multi-minute build+run", with `get_run_status` for polling.

I don't think there's a clean fix inside this PR, because `SimulationSet` has no non-blocking start (only `build`, `build_modified`, `run`). Two options:

- **Minimum for this PR:** keep it blocking, but document the starvation in the docstring so callers know what they're buying, and add the license pre-flight (other comment).
- **Follow-up:** run the batch on a dedicated thread outside the shared executor, return immediately, and add a sim-set status tool — mirroring the `run_project` + `get_run_status` pair.

Happy either way, but the current docstring reads as though this is routine, and it isn't.

PSCAD/pscad_mcp/tools/simset_tools.py
```diff
@@ -4,29 +4,53 @@
...
+
+    return (
+        f"Simulation set '{sim_set_name}' in project "
+        f"'{project_name}' completed."
+    )
```
qian-harvard, line 35: Two problems with this message:

- `project_name` is never resolved or validated, so this asserts a project association that was never checked. `run_simulation_set("WrongProj", "Batch1")` happily reports success "in project 'WrongProj'" for a set whose tasks belong to entirely different projects.
- "completed" isn't verified. `SimulationSet.run(consumer=None)` skips the build-events subscription entirely (`pscad.py:262-264` only subscribes `if handler`), so every task's compile can fail and this still reports completion.

Simplest version that doesn't overclaim:

```python
return f"Simulation set '{sim_set_name}' finished running."
```

If you want real status, `SimulationSet.list_tasks()` exists (`simset.py:192`), or you could pass a `consumer` to `run()` and collect the build events.

PSCAD/pscad_mcp/tools/simset_tools.py
```diff
@@ -4,29 +4,53 @@
 from pscad_mcp.core.executor import robust_executor
 
 async def list_simulation_sets(project_name: str) -> List[str]:
-    """List all simulation sets defined in a project."""
+    """List all simulation sets defined in the current workspace.
+
+    ``project_name`` is retained for backward-compatible MCP arguments.
+    Simulation sets belong to the PSCAD workspace, not to one Project.
+    """
+    del project_name  # Kept only for MCP schema compatibility.
```
qian-harvard, line 12: I don't think the schema-compatibility rationale holds up. The MCP SDK generates `inputSchema` from the function signature, marking any parameter without a default as required. Running these two signatures through the SDK's schema generator (`func_metadata`) gives:

```
run_simulation_set -> required: ['project_name', 'sim_set_name']
find_components    -> required: ['project_name']
```

(I had mcp 1.29 to hand rather than the pinned `mcp>=2,<3`, but this is core, stable schema-generation behavior — worth a sanity check on your end if you want to be sure.)

So the parameter isn't being preserved for compatibility; it's a required argument every client must invent and the tool then discards. And MCP clients re-fetch `tools/list` each session, so there's no older caller to stay compatible with.

`add_task_to_set` (line 44, same idiom) is the riskier one — it now has two project-shaped parameters side by side where the first is ignored, so a client that swaps them adds the wrong project as a task and gets no error.

I'd just drop the parameter from all three signatures (updating the two positional calls in the tests). If you'd rather keep it for now, `Optional[str] = None` at least stops it being required.

Good instinct on the docstrings, though — noting that "Simulation sets belong to the PSCAD workspace, not to one Project" is exactly right, and mcp surfaces it as the tool description.

qian-harvard (commented):
Thanks @Xiaoyang-Wang-TAMU, and apologies this sat for three weeks — its CI was never approved, so it had no checks and fell off the board. That is on us, not you.

The diagnosis is right and both bugs are real. I pulled the `mhi_pscad-3.1.2` and `mhi_common-3.3.0` wheels to check the vendor API rather than reason about it — they are `py3-none-any`, so they unpack and read fine off Windows.

## One blocker: two existing tests go green to red

Reproduced against your branch, with only a shim for the MCP SDK module my local `mcp` predates:

```
main    17 passed
pr-61    2 failed, 15 passed
```

```
FAILED PSCAD/tests/test_enhanced_tools.py::test_list_simulation_sets
  AssertionError: 'Batch1' not found in <MagicMock name='pscad_manager.pscad.simulation_sets()'>
FAILED PSCAD/tests/test_enhanced_tools.py::test_run_simulation_set
  AssertionError: 'started' not found in "Simulation set 'Batch1' in project 'TestProj' completed."
```

Both follow directly from the change and just need the tests moved with it:

1. `test_enhanced_tools.py:83` stubs `mock_project.simulation_sets`, but the tool now calls `pscad.simulation_sets`, so the stub no longer applies and a bare `MagicMock` comes back. The stub belongs on the workspace mock.
2. `test_run_simulation_set` asserts `"started"`; the tool now returns `"... completed."`. The new wording is more accurate — `run()` does block — so update the assertion, not the message.

Worth knowing these never ran in CI even when CI ran: `.github/workflows/test.yml:33` is `pytest tests/` only, so `PSCAD/tests/` is never collected. There are 67 test functions across `PSCAD/`, `HOPE/`, `PLEXOSDB/`, `PowerFactory/` and `surge/` in the same position. I am fixing the collection scope separately — not your problem to solve here.

## `find_all` — the fix is right, with one narrow gap

For the record, because I expected this to be worse than it is: collapsing both filters into `*criteria` is **correct** for the ordinary cases. `Project.find_all`'s own docstring (`mhi/pscad/project.py:1359-1397`) documents the signature as `find_all([[definition,] name,] ...)` and gives `find_all('Bus10')  # all components named "Bus10"`. So a lone positional is a *name*, and `find_all(definition, name)` with both is exactly right.

The gap is narrow and worth one guard. From the same docstring:

> If no definition was given, and if the provided name is "Bus", "TLine", "Cable", "GraphFrame", "Sticky", or contains a colon, it is treated as the definition name.

So `find_components(project_name="X", name="Bus")` — or any name containing a colon — silently becomes a definition search. Passing the name through the library's keyword path when no definition is given would close it.

Two smaller things in the same function:

- **`find_components(project_name="X")` with no filters now raises** (`project_tools.py:130-133`), but the generated MCP schema still marks only `project_name` as required, and the docstring is unchanged at line 116. "List every component in this project" is a natural first call for a model orienting itself, and it now gets a `ValueError` the schema said was legal. Either default to listing everything or make the schema reflect the constraint.
- **Per-component `parameters()` is N+1** (`:143-145`). `main` made one remote call and read attributes off the results; this adds a round-trip per match, each serialized through the single-worker executor with its own 30 s watchdog. Fine for a handful of matches, slow for a broad `definition=` sweep. Not a blocker — just flagging before someone runs it against a large case.

## Scope change is under-signposted

`list_simulation_sets` and `add_task_to_set` now `del project_name` and operate on the workspace. Your docstrings do say so, which is better than the diff suggested at first glance. But `project_name` is still a required positional with no default, so the MCP schema compels a caller to supply an argument that is immediately discarded — give it a default, or drop it and let the schema change. Relatedly, `run_simulation_set` still interpolates it into `"Simulation set 'X' in project 'Y' completed."`, which now names a project that had no bearing on what ran.

`PSCAD/README.md:115` still describes the old project scope.

## `_timeout=0` holds the only worker

`run_safe(sim_set.run, _timeout=0)` disables the watchdog while holding the lock on a `max_workers=1` pool, so a multi-hour batch blocks every other PSCAD tool with no recovery path if PSCAD freezes or throws up a modal dialog. I do not think there is a non-blocking alternative in the vendor API — `SimulationSet` has no `start()` — so this may be the only option, and `build_project` already does the same thing on `main`. Leaving it as a known trade rather than asking you to solve it, but worth a comment in the code saying so.

## Please strip the scratch markers

`project_tools.py:119-120` ships the previous implementation commented out, `:121` is `####### Code is modified`, and `:163` is a stray `########`. `PSCAD/pscad_mcp/` is inside the published `powermcp` distribution, so these land in users' site-packages. Git already records what changed.

## Tests

`git diff --name-only main...pr-61` shows no test files. The only existing `find_all` fixture returns `[]`, so none of the new code — `parameters()`, the `Name`/`name`/`NAME` fallback, `iid`, `defn_name` — is covered. One test where `find_all` returns a single mock component with those attributes would cover the whole new path.

---

To summarize what I am actually asking for: fix the two red tests, add a guard for the builtin/colon name case, drop the scratch markers, give `project_name` a default, and add one `find_all` fixture. The rest is discussion.

I am clearing the workflow approval on our side so you get real CI on your next push — the two runs from 28 August are still sitting at `action_required`, which is why this PR has never had a check.

One note on timing: `main` moved to `v0.3.1` an hour ago (#72) because the `powerio` pin was resolving to an incompatible version and `main` had been red since 6 September. Your branch is unaffected — it does not touch `powerio` — but rebase before your next push so CI runs against a green baseline.

PR #64, opened 2026-09-05
Modernize every bridge for PowerIO 0.11.2

samtalki:
PowerIO removed the 0.9 `Package` API in 0.11.0 and changed its default MCP path
policy in 0.11.2. This moves every PowerMCP bridge onto the typed PowerIO IR module
model, pins `powerio[mcp,matrix]>=0.11.2,<0.12` in all fourteen places, and returns
the suite to green.

## Solver boundary

`powermcp/solver_case.py` resolves one file or one serialized PowerIO IR document
into a typed module, selects a collection entry by `time_index` or `scenario_id`,
applies typed what-if edits through PowerIO's `OperatingPointUpdate`, `NetworkUpdate`
and `apply_bus_load_active_power`, and hands the solver a balanced network. It
re-parses and re-validates nothing PowerIO already states. Every adapter response
carries the same tail: `value_type`, `selection`, `diagnostics`, `warnings`, and the
emission `fidelity` where the adapter converts.

`pandapower`, `PyPSA`, `ANDES` and `Egret` take `powerio_ir` (with `network_json` as
an alias), `edits`, `to_balanced`, `base_mva`, `time_index` and `scenario_id`.
pandapower's hand-rolled PYPOWER-to-MATPOWER text projection is gone; PowerIO's
native writers do the conversion.

## Path containment

powerio 0.11.2 confines MCP paths to the directory the server process started in when
no root variable names one, where earlier versions left every path allowed. The
doctor now distinguishes an explicit root list from that implicit default, and the
test suite names the repository and pytest's temporary tree as its roots. This is the
root cause of the 41 failures on this branch's previous head.

## Also

- `surge` and `PyPSA` create the staging directory that powerio 0.11.2 now hands to a
  directory writer unset.
- `ANDES` and `Egret` install their emitted case file through powerio's staged file
  write, so a failed write leaves no truncated file behind.
- Collection selection rebuilds the module after every level, and a wrong `time_index`
  or `scenario_id` is refused with the available entries listed.
- Every adapter response, pandapower included, carries the `warnings` list.
- Four `print` calls that wrote to the MCP stdio channel move to stderr and logging
  (`ANDES`, `PowerWorld`).
- `GenX` stops leaking the repository root onto `sys.path` and declares its powerio
  dependency.
- `PLEXOSDB` stops citing a server file this repository deleted, and joins the SDK
  import sweep.
- CI drops the Rust toolchain step (powerio ships abi3 wheels) and cancels superseded
  pull request runs.
- Version 0.4.0.

## History

The change is the eight commits from `86dd105` to `d1a8d91`, rebuilt on current
`main`. The merge commit `064c0bc` adopts that tree; the branch's earlier commits
stay reachable through its second parent and contribute nothing to the diff.
Review the eight commits or the combined diff against `main`.

## Validation

- `pytest tests -q` on Python 3.12 against powerio 0.11.2:
  358 passed, 19 skipped, 0 failed.
- `tests/test_server_entrypoints.py` launches every registered server as a
  subprocess and lists its tools.
- `tests/test_transport.py` drives the powerio server over real stdio through
  `powermcp run powerio`.
- `python -m build` plus `twine check` on the sdist and wheel.

### DIFF
--- a/ANDES/andes_mcp.py
+++ b/ANDES/andes_mcp.py
@@ -16,7 +16,7 @@ _repo_root_added = _repo_root not in sys.path
 if _repo_root_added:
     sys.path.insert(0, _repo_root)
 try:
-    from powermcp.solver_case import resolve_solver_case
+    from powermcp.solver_case import resolve_solver_case, diagnostic_messages
     from powermcp.sandbox import (
         PathNotAllowed,
         checked_path,
@@ -497,8 +497,10 @@ def load_network_from_json(
     out_path: str,
     operating_point: Optional[int] = None,
     study_commit: Optional[int] = None,
+    time_index: Optional[int] = None,
+    scenario_id: Optional[str] = None,
 ) -> Dict[str, Any]:
-    """Stage PowerIO model JSON or one package state as MATPOWER for ANDES.
+    """Stage a selected PowerIO IR module as MATPOWER for ANDES.
 
     Accepts the ``json`` string returned by the powerio server's parse tool.
     Converts the network to MATPOWER format, writes it to out_path (use a .m
@@ -509,8 +511,10 @@ def load_network_from_json(
     Args:
         network_json: The JSON transport string from powerio
         out_path: Destination for the MATPOWER case file (.m)
-        operating_point: Optional package operating-point index to materialize
-        study_commit: Optional package study-commit index to materialize
+        operating_point: Compatibility alias for time_index
+        study_commit: Retired package selector; export a Tellegen Study state as IR
+        time_index: Explicit TimeSeries index
+        scenario_id: Explicit ScenarioSet identifier
 
     Returns:
         Dict with status, case_file path, component counts, and fidelity warnings
@@ -524,9 +528,11 @@ def load_network_from_json(
             network_json=network_json,
             operating_point=operating_point,
             study_commit=study_commit,
+            time_index=time_index,
+            scenario_id=scenario_id,
         )
         case = prepared.network
-        conv = case.to_format("matpower")
+        conv = prepared.emit("matpower")
         abs_out = os.path.abspath(out_path)
         with open(abs_out, "w") as fh:
             fh.write(conv.text)
@@ -541,7 +547,7 @@ def load_network_from_json(
             "branches": case.n_branches,
             "generators": case.n_gens,
         },
-        "warnings": list(prepared.warnings) + list(conv.warnings),
+        "warnings": list(prepared.warnings) + list(diagnostic_messages(conv.diagnostics)),
         **({"package": prepared.package} if prepared.package is not None else {}),
     }
 
@@ -553,6 +559,8 @@ def load_network_from_any(
     source_format: Optional[str] = None,
     operating_point: Optional[int] = None,
     study_commit: Optional[int] = None,
+    time_index: Optional[int] = None,
+    scenario_id: Optional[str] = None,
 ) -> Dict[str, Any]:
     """Stage any powerio readable case as a MATPOWER file for ANDES.
 
@@ -566,8 +574,10 @@ def load_network_from_any(
         out_path: Destination for the MATPOWER case file (.m)
         source_format: Input format name (matpower, powermodels-json, egret-json,
             psse, powerworld); inferred from the file extension when omitted
-        operating_point: Optional package operating-point index to materialize
-        study_commit: Optional package study-commit index to materialize
+        operating_point: Compatibility alias for time_index
+        study_commit: Retired package selector; export a Tellegen Study state as IR
+        time_index: Explicit TimeSeries index
+        scenario_id: Explicit ScenarioSet identifier
 
     Returns:
         Dict with status, case_file path, component counts, and fidelity warnings
@@ -586,9 +596,11 @@ def load_network_from_any(
             source_format=source_format,
             operating_point=operating_point,
             study_commit=study_commit,
+            time_index=time_index,
+            scenario_id=scenario_id,
         )
         case = prepared.network
-        conv = case.to_format("matpower")
+        conv = prepared.emit("matpower")
         abs_out = os.path.abspath(out_path)
         with open(abs_out, "w") as fh:
             fh.write(conv.text)
@@ -605,7 +617,7 @@ def load_network_from_any(
             "branches": case.n_branches,
             "generators": case.n_gens,
         },
-        "warnings": list(prepared.warnings) + list(conv.warnings),
+        "warnings": list(prepared.warnings) + list(diagnostic_messages(conv.diagnostics)),
         **({"package": prepared.package} if prepared.package is not None else {}),
     }
 
--- a/Egret/egret_mcp.py
+++ b/Egret/egret_mcp.py
@@ -16,7 +16,7 @@ _repo_root_added = _repo_root not in sys.path
 if _repo_root_added:
     sys.path.insert(0, _repo_root)
 try:
-    from powermcp.solver_case import resolve_solver_case
+    from powermcp.solver_case import resolve_solver_case, diagnostic_messages
     from powermcp.sandbox import PathNotAllowed, checked_path, ensure_checked_directory
 finally:
     if _repo_root_added:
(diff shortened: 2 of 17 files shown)


qian-harvard (commented):
Reviewed at depth. The substance here is sound and unusually well tested — `solver_case.py` is a real improvement over the hand-rolled projection it replaces, and the typed-edit boundary is the right shape. I want this in. Three things first, plus a decision about this PR versus #71 that I think needs settling before either moves.

## This PR, not #71

I checked whether #64 is still needed given #71, because on the surface #71 looks like a strict superset. It is:

```
$ git diff pr-71...pr-64 --stat
(empty)
$ comm -12 <(git diff --name-only main...pr-64|sort) <(git diff --name-only main...pr-71|sort) | wc -l
52          # all 52 of #64's files
```

So #71 contains every line of this PR. But it also bundles #67 (the Tellegen server) and #70 (the tool-result-shape change) — two drafts that have not been reviewed — and merging #71 today would land all four PRs in one click. That is not a trade I want to make for a 2,000-line dependency migration.

It also carries a test that fails. `PLEXOSDB/plexosdb_mcp/main.py:183` returns `tool_success(...)` on #71, and `PLEXOSDB/tests/test_tools.py:129` still asserts the old shape:

```python
>>> tool_success(output_path="/tmp/out/system.json", model_name="Base", component_types={"Bus": 2})
{'status': 'success', 'output_path': '/tmp/out/system.json', 'model_name': 'Base', 'component_types': {'Bus': 2}}
>>> result["ok"]
KeyError: 'ok'
```

CI cannot see it: `.github/workflows/test.yml:33` runs `pytest tests/` only, so `PLEXOSDB/tests/` is never collected. That belongs to #70 and I will raise it there — noting it here only because it is the reason I am not taking the superset.

**So: this PR is the one to merge.** #67, #69, #70 and #71 should then rebase onto the result, at which point their real diffs are small (#69 is 4 files, #71 is 23 of which 16 are one-token version bumps).

## You will need a rebase, and that one is on me

I merged #72 an hour ago to cap `powerio` at `<0.10`. `main` had been red since 6 September — the `<1` ceiling was resolving to 0.11.x against 0.9-era code, so `pip install powermcp==0.3.0` was broken for everyone and CI was running 61 failed / 283 passed. That is now `v0.3.1`, 344 passed.

It collides with this PR in exactly the 15 files it touched — the 13 `powerio` pin sites plus `powermcp/__init__.py` and `tests/test_registry.py`. Every hunk is a one-line version string where both sides edited the same line, so the resolution is "take yours" throughout: your `>=0.11.2,<0.12` supersedes my `<0.10` cap, and your `0.4.0` supersedes `0.3.1`. Sorry for the churn — the alternative was leaving a broken package on PyPI for however long this PR takes.

Please also re-run CI after rebasing. The last green run installed powerio 0.11.2, and 0.11.3 went up on 17 September, so the current pin now resolves to a version this branch has never been tested against. I checked the 0.11.2→0.11.3 delta myself and it is additive everywhere PowerMCP touches — `allowed_roots()` is byte-identical, the sandbox diff is comment-only — so I expect it to be clean. Worth the click anyway.

## Findings

**1. The multiconductor guard is unreachable for four of the five types it covers** — `powermcp/solver_case.py:478-483`

```python
if isinstance(value, _MULTICONDUCTOR):          # 5 types (:399-405)
    if not to_balanced or not isinstance(value, powerio.MulticonductorNetwork):
        raise ValueError("... or pass to_balanced=True")
```

`_MULTICONDUCTOR` is `(MulticonductorNetwork, McAcPfInstance, McAcOpfInstance, McAcPfSolution, McAcOpfSolution)`. For the four that are not `MulticonductorNetwork`, the second clause is true regardless of `to_balanced`, so the raise fires unconditionally — and tells the caller to pass `to_balanced=True`, which is what they just did. A serialized `McAcPfInstance` is a dead end with advice that cannot work.

If lowering genuinely only applies to a network and not to a calculation instance, the message should say that instead of naming a flag. If instances should be lowerable, the guard is wrong.

**2. `powermcp doctor` names its own startup directory as the servers' sandbox** — `powermcp/doctor.py:145-158`

The policy description is right and the docstring is honest about it. The problem is `{listed}`: `allowed_roots()` runs in the doctor process, and powerio captures the startup directory at import, so the reported root is wherever `powermcp doctor` was invoked from. Servers launch as separate processes from the MCP client's working directory, which is usually somewhere else. An operator who runs the doctor from the repo, sees a plausible-looking path, and moves on has been told something that is not true of the thing they are about to run. Either name the directory the servers will actually start in, or drop the specific path and describe the rule.

**3. The containment change needs a migration note, and `wizard.py` is where it should land**

`powermcp/sandbox.py:14-15` — powerio 0.11.2 flips unset roots from "nothing is constrained" to "confined to the server's startup directory". Every existing 0.3.x deployment that never set `POWERIO_MCP_ALLOWED_ROOTS` silently narrows on upgrade, and will surface as `PathNotAllowed` on paths that worked yesterday. The README prose covers it, but `wizard.py`'s diff is 10 lines of pure punctuation — it still never prompts for roots, and the generated client entries carry no `env` block. At minimum this needs a line in the release notes; ideally the wizard asks.

**Minor:** the PR body says "`GenX` stops leaking the repository root onto `sys.path`", but `GenX/server.py:22-24` still does `sys.path.insert(0, _REPO_ROOT)` exactly as `main` does — the change is three comment lines explaining why it stays. Worth fixing the description so a reviewer does not go looking for a change that is not there.

## Explicitly dropping

I chased two concerns to the powerio source and they do not hold — flagging so nobody re-raises them:

- The private `_inner._type_name` read is **not** at risk from 0.11.3. The 0.11.2→0.11.3 diff of `python/powerio/__init__.py` removes exactly one line and it is a docstring; `_type_name` is still declared at `_powerio.pyi:879`.
- `IR_VERSION` being unused is real but the `document["version"]` KeyError it implies is unreachable: `StoredModule` declares `version: u64` with `deny_unknown_fields` and no `serde(default)`, so a document without it fails to deserialize before that subscript runs. Dead constant, not a bug.

Also worth saying: `tests/test_publish_workflow.py` plus the new sdist excludes for `.venv*` and `.uv-build-cache` are a genuine packaging fix — hatchling would otherwise ship a developer virtualenv to PyPI — and it is not mentioned in the description at all. It deserves a line.

---

Rebase, re-run CI, and answer 1. Then I will merge this and we can work through the stack on top of it.

qian-harvard:
Status check on this stack (#64 → #67 → #70 → #71, with #69 also on #67): all five now conflict with `main`, and it's worth saying exactly how, because the fix is mechanical and the direction is not in dispute.

#72 landed on `main` after this branch was cut and capped the pin at `powerio[mcp,matrix]>=0.9.0,<0.10`, as a stopgap so a fresh install resolves something usable. This PR moves it to `>=0.11.2,<0.12`. So every conflict is the same conflict, repeated across the fourteen pin sites:

```
ANDES/  Egret/  LTSpice/  OpenDSS/  PSLF/  PSSE/  PowerFactory/
PowerWorld/  PyPSA/  pandapower/  surge/   (requirements.txt)
PSCAD/pyproject.toml   pyproject.toml   powermcp/__init__.py   tests/test_registry.py
```

`git merge-tree main` reports no conflicts outside that set for this PR. #71 adds one more, `PowerFactory/MCP_PowerFactory.py`, from #66. Resolution is to take the branch side on the pins — the `<0.10` cap was always meant to be temporary, and this stack is what replaces it. `powermcp/__init__.py` needs the version to stay in step with `pyproject.toml` and `tests/test_registry.py` or `test_package_versions_match` fails.

One thing that has changed in this stack's favour: **powerio 0.11.3 is on PyPI now** (`pip index versions powerio` → latest 0.11.3). #71 says it's "draft until powerio 0.11.3 is on PyPI: until then the pin cannot resolve and CI cannot install the package" — that blocker is gone, and #71 is already marked ready for review. If the plan is still to land the stack bottom-up, this one needs the rebase first; if 0.11.3 makes #64's 0.11.2 floor redundant, collapsing the stack onto the 0.11.3 pin may be less work than rebasing four branches in order.

Either way I'd rather not resolve fourteen pin conflicts on your behalf and guess at the intent — could you rebase, and say which way you want to sequence it?

Issue #10, opened 2025-11-18
fix(docs): Correct broken 'Build MCP server' link


martymcenroe:
The README.md file contains a broken link for 'Building Your First MCP Server'. \n\n* **Broken:**  (leads to 404)\n* **Correct:** \n\nThis commit updates the link to the correct URL.

qian-harvard:
Thanks for the comments. The error is fixed now.

PR #66, opened 2026-09-09
Fix PowerFactory follow-up review findings

### LINKED ISSUE #65: PowerFactory: follow-up review findings from #63
Follow-up to #63 (merged as 97e3dd9). That PR is a large, well-structured addition and nothing in it regresses existing tools, so it was merged as-is. This issue collects the review findings so they can be addressed separately.

Note on verification: these were found by reading the code, running the PR's 12 unit tests, and cross-checking the committed evidence transcripts. **No live PowerFactory was available**, so items marked _(unverified against live PF)_ are reasoned from code and may turn out not to apply.

## Correctness

**1. `delete_component` reports `deleted: false` after the component was actually deleted** — `MCP_PowerFactory.py:712`

When `update_graphics=true` and a graphic or cubicle cleanup step fails, `Agent_DIgSILENT.py:1799` raises *after* `component.Delete()` already succeeded, so the wrapper computes `deleted = ok and bool(confirmation)` = `false`. `app.Rebuild()` is also skipped, leaving a stale diagram. The PR's own test (`test_component_creation.py:815-818`) asserts `ok is False` while `grid["ElmLod"] == []`. A client trusting `deleted: false` retries and gets "Load not found".

Suggested fix: return a structured result (`deleted`, `graphics`) from the agent rather than deriving `deleted` from the request.

**2. The six read-only tools never connect, so they fail on a fresh server** — `MCP_PowerFactory.py:227` (also 253, 283, 385, 446, 496)

They read `DIgSILENTAgent._shared_app` and return "PowerFactory is not connected" when it is `None`, but no read-only path ever sets it. Every write to `_shared_app` lives in `Agent_DIgSILENT.py`; the MCP layer only reads it. So `get_active_project()` on a fresh server fails even with PowerFactory open on a project, and only starts working after a *mutating* tool has run. The README's own example prompt ("Return the active PowerFactory project and study case ... without modifying anything") cannot work cold.

Suggested fix: route all six through `DIgSILENTAgent._get_application(open_digsilent=False)` in one shared accessor. Related: once connected, PF errors inside these `_impl` bodies are unguarded and escape as raw tool exceptions instead of the `{"success": false, ...}` envelope every other tool returns.

**3. PowerFactory's 40-character `loc_name` limit breaks rollback verification** — `Agent_DIgSILENT.py:1206`

The committed evidence shows the truncation: `FT-11` line 30, `FT-06` line 32, and `FT-07` line 31 all show generated cubicle names cut to exactly 40 characters (in FT-07 both transformer cubicles collapse to the *same* truncated name).

- `_rollback_connected_element` (1143-1150) compares the truncated stored name against the untruncated `f"{element_name} Cubicle"`, so for element names of 33+ characters `cubicle_exists` is structurally `False` and the tool reports `rolled_back=True` even when `cubicle.Delete()` failed and a dangling cubicle + breaker remains.
- A 41+ character component name passes validation, creates cubicles and switches, then fails at line 1286 with the misleading "PowerFactory did not retain the load name".

Suggested fix: validate name length up front and compare against the actual stored `loc_name`.

**4. `delete_component` removes any connected cubicle wholesale** — `Agent_DIgSILENT.py:1772`

Lines 1705-1712 collect the cubicles referenced by `bus1`/`bus2`/`bushv`/`buslv` with no origin or content check, and 1770-1774 delete them inside `except Exception: pass`. For a pre-existing model element, those cubicles may contain `ElmRelay`, `StaCt`/`StaVt`, or be the target of an `EvtSwitch`. README lines 61 and 158-159 promise only "generated cubicles" are removed, and the preview (1716) shows just the component's full name, so the user confirms without seeing the collateral. _(unverified against live PF)_

Suggested fix: list the cubicles and their contents in the preview, or only delete cubicles that are empty apart from the generated `StaSwitch`.

**5. `get_parameters` crashes on object-list attributes** — `MCP_PowerFactory.py:316`

`list` is whitelisted without inspecting its elements, and `_to_json._clean` (162-179) returns unknown objects unchanged, so a reference-vector attribute makes `json.dumps` raise `TypeError` out of the tool instead of landing in the per-variable `errors` map the code otherwise builds carefully. Reproduced with a dummy object inside a list.

Suggested fix: stringify non-primitive list elements recursively, or add a `str()` fallback in `_clean`.

**6. `add_component` matches names case-insensitively, `delete_component` exactly** — `Agent_DIgSILENT.py:1683`

With a grid containing `bus a`: `add_component("bus", "Bus A")` returns "Bus already exists" (casefold, line 1177), but `delete_component("bus", "Bus A", ...)` returns "Bus not found" (exact, line 1683). Neither message surfaces the real stored name. Reproduced with the test fakes.

**7. Parameter coercion edge cases** — `Agent_DIgSILENT.py:1474` and `:1467`

All reproduced with the test fakes:
- `"reactive_power_mvar": null` is rejected with "must be a number", although the docstring lists it as optional and `dict[str, Any]` lets JSON null through. Omitting the key works.
- `"active_power_mw": true` silently creates a 1.0 MW load; `"nominal_voltage_kv": true` creates a 1.0 kV bus.
- A bus named `"0"` passed as the integer `0` gives "bus_name must not be empty", because `str(parameters[key] or "")` evaluates `0 or ""` first.

**8. Diagram Layout Tool: dead guard and unlinked start set** — `Agent_DIgSILENT.py:1323`

`GetFromStudyCase(name)` creates the object when absent, so the "start-element set is unavailable" guard at 1325-1328 can never fire (the no-study-case case is already caught at 1318-1320). The set is found only by the GUI-generated name `Set - SGL Layout - K-neighbourhood` and is never linked to the `ComSgllayout` command, so on a study case where the layout tool was never run interactively — or a localized install — the code may fill a fresh, unlinked set, `Execute()` inserts nothing, and `add_component` returns `success=false` for an element that now exists (a retry then hits "already exists"). The evidence only covers runs on a pre-existing developer study case, and the unit test hard-codes the set name into the fake. _(dead guard confirmed; unlinked-set consequence unverified against live PF)_

**9. Delete confirmation token is not bound to the grid** — `Agent_DIgSILENT.py:1713`

The token is `f"DELETE {kind} {name}"` and excludes `grid_name` and the previewed full path. With two active grids each containing `Load 1`, a token obtained by previewing in Grid A confirms deletion in Grid B with `success=true`. The docstring's promise of "the exact confirmation token returned by the preview" does not protect the object that was previewed.

Suggested fix: include the full name (or grid) in the token.

**10. `finally` restore can mask the real error; `Unfreeze` never reverted** — `Agent_DIgSILENT.py:1347`

`start_elements.Clear()` / `AddRef(existing)` in the `finally` can themselves raise, replacing the original `Execute()` diagnostic, or turn a successful insertion into "graphical update failed". Both reproduced with mocks; the trigger needs a set member to become invalid between lines 1330 and 1348, so it is rare. Separately, `desktop.Unfreeze()` (1336, 1748) has no matching `Freeze` anywhere in the file and the prior state is never saved, so every `update_graphics=true` call permanently leaves the user's diagram unfrozen.

**11. Orphan cubicles create a dead end** — `Agent_DIgSILENT.py:1700`

The bus guard rejects any bus with a `StaCubic` child rather than only cubicles whose `obj_id` points at a live element. If a late failure leaves an orphan cubicle (the rollback swallows `Delete()` errors), then re-adding gives "Cubicle already exists on bus", deleting the component gives "not found", and deleting the bus gives "Bus has connected cubicles; delete its connected components first" — although nothing is connected. Recovery requires the PowerFactory GUI. _(unverified against live PF)_

**12. `labels[attribute]` raises `KeyError` on the failure path** — `Agent_DIgSILENT.py:1108`

`_set_and_verify_attributes` accepts an arbitrary `expected` dict but indexes a hard-coded eight-entry `labels` dict when raising. Latent today (all current attributes are covered), but the first new attribute anyone adds turns a "did not retain" error into a bare `KeyError`. One-line fix: `labels.get(attribute, attribute)`.

## Documentation

**13. README lists six agent methods that do not exist** — `PowerFactory/README.md:87-92`

`DIgSILENTAgent.get_active_project`, `get_active_study_case`, `get_parameters`, `list_objects`, `list_components`, `list_study_cases` are documented as agent methods, but none are defined on the class — they exist only as inline closures in the MCP handlers. `functions_overview.txt` correctly lists only `add_component`/`delete_component`, so the two docs disagree. Calling them from a script gives `AttributeError`.

Also: the `delete_component` docstring says it deletes from "the currently active single-line diagram", but `_find_component_graphics` walks every `IntGrfnet` in the project and deletes matches from all of them. Deleting from all diagrams is arguably correct (no orphan symbols), so this is a wording fix.

Minor: the section header at `Agent_DIgSILENT.py:1834` reads `# # LOAD FLOW` (doubled `#`); the header at 950 (`# ADD BUS - create a verified ElmTerm`) labels ~880 lines that also contain the deletion and diagram logic; and the `*` marker on eight tools in the module docstring has no legend.

## Packaging / repo hygiene

**14. Evidence and reports ship in the PyPI wheel, including an account identifier**

`pyproject.toml` force-includes the whole `PowerFactory` directory into the wheel, and the sdist `exclude` list only strips `PSCAD/tests`, `HOPE/tests`, `PLEXOSDB/tests`. So ~1,300 lines of report markdown, 13 evidence transcripts, and the unit tests are all packaged and published.

The evidence transcripts and `POWERFACTORY_MCP_CONTRIBUTION_REPORT.md:68` also contain the contributor's real PowerFactory account identifier (19 occurrences across nine files). `evidence/README.md:9` states "The main report uses anonymized paths for presentation", which holds for the function-test report but not for the contribution report. No other connector on `main` commits evidence transcripts or `*REPORT*.md` files.

Suggested fix: anonymize the account identifier and exclude `evidence/`, the report files, and tests from the wheel.

**15. The PR's tests never run in CI** — `PowerFactory/test_state_inspection.py:30`

`.github/workflows/test.yml` runs only `python -m pytest tests/ -q`, which collects zero tests from `PowerFactory/test_*.py`, so all 12 new tests are dead weight in CI. Two problems in the test module itself:

- It replaces `sys.modules["Agent_DIgSILENT"]` with a bare fake at import time and never restores it. Loading `test_state_inspection` before `test_component_creation` makes 10 of the 12 tests error with `AttributeError: module 'Agent_DIgSILENT' has no attribute 'pf'`. It passes today only because alphabetical order happens to be favourable — which is the order the `UT-01` evidence recorded.
- It stubs `sys.modules["fastmcp"]`, but `MCP_PowerFactory.py:64` imports `from mcp.server.mcpserver import MCPServer as FastMCP`, so the stub is dead code and the module still requires that exact SDK.

Suggested fix: scope the stubs with `patch.dict(sys.modules, ...)` in `setUpModule`, stub the module that is actually imported, and wire these tests into `tests/` so CI runs them.

## Cleanup (lower priority)

- `_get_application` (`Agent_DIgSILENT.py:1013`) was added but the six pre-existing inline copies of the same connect-or-reuse block (`connect`, `import_project`, `modify_parameter`, `load_flow`, `short_circuit`, `create_study_case`) were left untouched, so connection logic now lives in seven places. A mechanical `app = cls._get_application(open_digsilent)` swap is behaviour-preserving in each.
- Component-kind knowledge is encoded in four places that must stay in sync by hand — `schemas` (1407), the attribute ladder (1489-1545), the summary ladder (1589-1630), and `component_types` in `delete_component` (1655) — plus a fifth copy in the MCP docstring. Adding a kind to one and not the others yields objects that can be created but not deleted. A single spec table would drive all of them.
- `_agent_result` (`MCP_PowerFactory.py:138`) has one caller; four older tools are exact-shape matches for it, and `delete_component` inlines the same pattern.
- `list_components` reads four attributes for *every* matching object before truncating to `max_results` (unlike `list_objects` and `get_parameters`, which slice first), and issues 14 separate `GetCalcRelevantObjects` calls for `"all"`.
- Several lookups scan a whole class and call `GetAttribute("loc_name")` per object to find one name; `delete_component` repeats the full scan after `Delete()` to confirm removal.

---

Reviewed with `/code-review` at extra-high effort. Happy to split any of these into separate issues if that is easier to work through.

aswinkrishnapoyil:
## Summary

Addresses the PowerFactory follow-up findings reported in #65.

- Return structured deletion results that accurately report partial deletion and graphical cleanup outcomes.
- Connect read-only tools on a cold server and return consistent error envelopes.
- Validate PowerFactory name limits and use retained object names during rollback.
- Preserve protected cubicles and allow safe cleanup of orphan cubicles.
- Handle object-list parameters, case-insensitive names, null optional values, booleans, and numeric bus names safely.
- Use automatic diagram insertion without K-neighbourhood relayout, restore the desktop after creation, and leave desktop state untouched during deletion.
- Bind deletion confirmation tokens to the component's full PowerFactory path.
- Correct API documentation and anonymize verification artifacts.
- Exclude reports, evidence, and PowerFactory tests from distributed packages.
- Run the PowerFactory unit tests in CI.
- Centralize application access, component specifications, and result wrapping.
- Optimize component listing and use targeted object lookups instead of whole-class scans.

## Verification

- `git diff origin/main...HEAD --check`: passed
- PowerFactory unit tests: `26 passed`
- Package build: wheel and sdist built successfully
- Package-content verification: no evidence, reports, or PowerFactory test files included
- Privacy scan: no contributor PowerFactory account identifier remains in committed files
- Live PowerFactory targeted-lookup smoke test: all 10 checks passed
- Temporary live-test load and cubicle were successfully removed

Closes #65


qian-harvard (commented):
Thanks for this, @aswinkrishnapoyil — and sorry it sat for nine days. This is a thorough piece of work and it addresses the great majority of #65 properly rather than superficially. I spot-checked several findings against the diff rather than taking the summary on trust:

- **#2** — all six read-only tools now route through `_read_only_result` → `_get_application(open_digsilent=False)` (`MCP_PowerFactory.py:156-159`), so the cold-server path works. Confirmed at all six call sites.
- **#9** — the confirmation token is now `f"DELETE {kind} {component.GetFullName()}"` (`Agent_DIgSILENT.py:1763`), so it is bound to the full PowerFactory path and no longer transfers between grids.
- **#14** — `evidence/`, `POWERFACTORY_MCP_*_REPORT.md` and `PowerFactory/test_*.py` are all excluded from the distribution now (`pyproject.toml:162-163`).

The `_component_specs` table and the `_get_application` consolidation also close out most of the cleanup section, which I had marked lower priority — nice to see them done.

Two things I cannot check without a live PowerFactory, and one I would like a second opinion on before this goes in.

---

### 1. Possible case-sensitivity regression in `_find_named_contents` — finding #6 may be inverted, not fixed

This is the one I would most like you to confirm, because if I am right it turns finding #6 from "inconsistent" into "both paths broken."

`_find_named_contents` (`Agent_DIgSILENT.py:1008-1017`) moved the name into PowerFactory's own query string:

```python
parent.GetContents(f"{requested}.{class_name}", 1) or []
```

whereas `main` queried with a wildcard and did all name matching in Python:

```python
grid.GetContents(f"*.{class_name}", 1) or []      # main:1139
...
str(obj.GetAttribute("loc_name")).casefold() == element_name.casefold()   # main:1142-43
```

The `casefold()` comparison is still there in the new version, but it can now only filter *down* from whatever PowerFactory already returned. **If PF's `GetContents` pattern matcher is case-sensitive**, then `_find_named_contents(grid, "Bus A", "ElmTerm")` returns `[]` when the stored name is `bus a` — the casefold check never gets a row to match. `add_component` would then stop detecting the duplicate and silently create a second `Bus A` alongside `bus a`, which is worse than the inconsistent-error-message behaviour #6 described.

The unit tests can't catch this either way, because the fake discards the name half of the query (`test_component_creation.py:73`):

```python
def GetContents(self, query, recursive):
    self.content_queries.append(query)
    if query == "*":
        return [...]
    return list(self[query.rsplit(".", 1)[-1]])     # <- "Bus A.ElmTerm" -> "ElmTerm"
```

So the fake always returns every object of the class and lets the Python `casefold()` do the work — i.e. it models `main`'s wildcard behaviour, not the new query.

**Could you check on your live install** whether `GetContents("Bus A.ElmTerm", 1)` finds an object stored as `bus a`? If it is case-sensitive, the fix is to keep the wildcard pre-query and leave the matching in Python. Either way it would be worth teaching the fake to honour the name half of the query, so the test actually exercises what the code now does.

### 2. Desktop freeze state after an add + delete cycle — finding #10

`_update_active_diagram` (`:1413-1414`) sets a fixed freeze state rather than restoring the observed one. The add path looks self-healing since it unfreezes at `:1416` on every call, so as far as I can tell the only exposed path is `delete_component(update_graphics=true)` reaching `app.Rebuild()` at `:1860`. Can you run one add + delete cycle and confirm the user's diagram is not left frozen?

### 3. Minor — `evidence/UT-01_automated_tests.txt`

The record is dated 1 Sept but names a test written on the 9th, and still reports "Ran 12 tests" for what is now a 26-test suite. Worth re-capturing or dropping the date, since these transcripts are the verification record for work no reviewer here can reproduce.

---

None of this is a reason to hold the PR long. Answer 1 and 2 and I will merge. Given that #70 is going to rewrite `_agent_result` into a different result shape and collides with this PR across 15 hunks in `MCP_PowerFactory.py`, I want your version to land first — the semantics here were worked out against a real engine, and that is the harder half to reconstruct.

One heads-up unrelated to your changes: `main` itself has been red since 6 Sept because the `powerio` pin admits 0.11.x against 0.9-era code (`main` runs 61 failed / 283 passed). #72 caps it and is green at 344 passed. Your CI here is green because your branch predates the resolver picking up 0.11 — so please hold off rebasing until #72 lands, or you will pick up 61 unrelated failures.

qian-harvard:
Merged. Thanks @aswinkrishnapoyil — this was a lot of careful work against findings that were tedious to chase, and it closes #65.

I merged on your verification rather than waiting for answers to my two questions. Neither is checkable by anyone here — no maintainer on this repo has a PowerFactory install — and holding a green, well-tested PR for a live check only you can run was the wrong trade. The desktop-freeze question I am content to leave; the `GetContents` case-sensitivity one is tracked in #73 so it does not get lost. No action needed unless the live check comes back empty.

One thing I should flag since it touched your PR: I resolved a conflict in `pyproject.toml` by hand while merging. You and #72 independently capped `powerio` at `<0.10` — same pin, different comment — so the resolution kept your one-line summary plus the concrete detail from #72. Nothing of yours was dropped. Your pin hunk was the right call and it would have fixed the broken release on its own had this landed sooner; sorry again for the nine-day wait that made #72 necessary.

Also worth saying: your PR now runs the PowerFactory tests in CI, so the 26 tests are no longer dead weight. That is a real improvement on the `pytest tests/` blind spot and I am extending the same treatment to the other connector suites separately.

aswinkrishnapoyil:
Thanks for the review and for merging #66. Following up on the three points you raised:

1. **GetContents case sensitivity:** Your concern was correct. On PowerFactory 2026 SP1, querying the stored `Bus 01` using `Bus 01.ElmTerm` returned one object, while `bus 01.ElmTerm`, `BUS 01.ElmTerm`, and `bUS 01.ElmTerm` returned none.

   I have pushed a fix for #73 that restores the class wildcard query and performs exact case-insensitive matching in Python. The test fake now honours case-sensitive name patterns, with regression coverage for duplicate rejection, bus selection, and deletion lookup.

   All 29 tests and 3 capitalization subtests pass. A live check also confirmed that the corrected helper returns the same existing bus for all four capitalization variants.

2. **Desktop freeze state:** I understand from your merge comment that you are content to leave this question. Our earlier screenshots checked diagram movement, but they did not establish whether the previous freeze state was restored. I therefore cannot claim that specific behaviour has been verified.

3. **Automated-test evidence:** You are right that the committed transcript has an inconsistent date and outdated test count. The latest run for the #73 fix reports 29 tests and 3 subtests passing. The historical transcript has not yet been corrected; I will address that separately rather than present it as current evidence.

Thank you for identifying the query behaviour and the gap in the test fake—the live check confirmed both were important.

The fix and live verification results are available in #77.

PR #67, opened 2026-09-16
Add a Tellegen MCP server over the native CLI

samtalki:
**Stacked on #64.** This branch starts from `codex/powerio-011-modules`, so its diff
against `main` includes that pull request until it merges. Review only the commits
after #64's tip, or wait for #64 to land.

`powermcp/tellegen.py` runs the native [Tellegen](https://github.com/eigenergy/tellegen)
CLI as an MCP server. It exchanges PowerIO IR only: the powerio server parses, Tellegen
solves, and the solution comes back as a PowerIO module every other tool reads.

## Tools

`capabilities`, `contract`, `solve`, `solve_module`, `plan`, and the Study family
`study_contract`, `study_create`, `study_inspect`, `study_run`, `study_export`,
`study_import`. `study_run` takes one of eight operations. Applying a Study proposal
binds a recommendation to the Study and stays a human action: it is not a tool, and
`study_run` refuses it.

## Behavior

- The CLI resolves from `POWERMCP_TELLEGEN_BINARY`, then the `tellegen.binary` config
  key, then `tellegen` on `PATH`.
- Requests go over stdin as JSON, one JSON object comes back on stdout, and progress
  events arrive one JSON object per line on stderr.
- `POWERMCP_TELLEGEN_TIMEOUT_SECONDS` defaults to 1800 and
  `POWERMCP_TELLEGEN_CANCEL_GRACE_SECONDS` to 300. Cancellation sends SIGTERM, or
  CTRL_BREAK to a Windows process group, so the current exact trial finishes and its
  evidence is saved before a forced stop.
- A multiconductor value is refused before the binary runs, naming the powerio
  server's `to_balanced` as the remedy. A module PowerIO marks with an error is
  refused on the same terms.
- The extra is empty by design: the CLI is a Rust binary, not a pip package.

## Validation

- `pytest tests -q`: 390 passed, 20 skipped, 0 failed
  (Python 3.12, powerio 0.11.2).
- 30 of the 31 Tellegen tests run against `tests/data/fake_tellegen.py`, a Python
  stand-in for the compiled CLI, so CI needs no Rust toolchain.
- `TELLEGEN_BIN=<path> pytest tests/test_tellegen_server.py`: 31 passed against a
  `cargo build -p tellegen-cli --features conic` build of eigenergy/tellegen `3130e92`
  (powerio crates 0.11.1), including the real-binary case9 DC OPF solve; the doctor
  reports the configured binary as ok.
- `tests/test_server_entrypoints.py` covers `tellegen` automatically through the
  registry.


qian-harvard:
#64 is merged (`809be61`), so this branch's stated reason for being a draft — "Review only the commits after #64's tip, or wait for #64 to land" — is resolved. Against the new `main` this now merges **clean**, and I ran the suite on the merge result in a clean 3.12 venv: **393 passed, 20 skipped**. Its real diff is down to 31 files, +1909/-1218, with `powermcp/tellegen.py`, `powermcp/TELLEGEN.md`, `tests/test_tellegen_server.py` and `tests/data/fake_tellegen.py` as the new files.

I'm not going to flip your draft and merge it, though, because this isn't a conflict question any more — it's a product one. It adds a whole MCP server with an external Rust binary as its engine (`POWERMCP_TELLEGEN_BINARY` / `tellegen.binary` / `PATH`), and an empty extra by design, so `pip install powermcp` gives users a server they can't run until they've built eigenergy/tellegen themselves. That's a reasonable design for a native solver, and the fake-CLI test strategy means CI stays honest — but whether the bundle takes it on is yours and @qian-harvard's call, not something I should decide by merging.

Two things I'd want settled before it lands:

1. **Discoverability of the unusable state.** If the binary isn't present, what does a model calling `capabilities` see? A clear "not installed, build it from <url>" beats a generic failure, given the server will be listed by `powermcp` regardless.
2. **The 3.13/3.14 matrix.** I cherry-picked #69's CI commit onto `main` (`19c1e9b`), so this will now be exercised on four interpreters instead of two. Worth a glance that the subprocess and signal handling (`CTRL_BREAK` on Windows process groups) is fine there — your validation was 3.12 only.

Mark it ready when you're happy and it can go in; the merge itself is a non-event now.

PR #68, opened 2026-09-16
feat: add deterministic network audit layer

BurhanAbdullah:
Adds a solver-independent structural audit layer for pandapower-compatible networks.

The audit provides deterministic, machine-readable findings for common pre-solve network issues, including:

invalid bus voltage limits
out-of-service buses
invalid line length/impedance/rating
invalid transformer ratings
disconnected in-service buses

The audit does not execute a power flow and does not mutate the input network.

Includes regression tests covering clean networks, invalid voltage limits, invalid line parameters, and the no-power-flow contract.

This is intentionally isolated from solver execution so MCP/agent workflows can perform a deterministic validation step before invoking a solver.


qian-harvard (commented):
Thanks for this — the idea is sound. A solver-independent pre-flight check is genuinely useful for agent workflows, where the alternative is handing a model a raw `LoadflowNotConverged` and hoping it can reason backwards to the cause. The structural checks read cleanly, the report shape is stable and JSON-serializable, and I like that `to_dict()` is tested for determinism.

I ran it against real networks before reviewing the diff. Two things need to change before it can merge.

## Blocking: `DISCONNECTED_BUS` can never fire

`audit.py:126-131`:

```python
connected = set().union(*(set(component) for component in nx.connected_components(graph)))
...
for idx in sorted(in_service_buses - connected):
```

The union of *all* connected components is, by definition, every node in the graph. So `connected` is always the full node set, `in_service_buses - connected` is always empty, and the check emits nothing — ever. It doesn't detect islanding, and it doesn't detect a bus with no branches at all.

Reproduced on your branch (`pandapower` 3.x, `networkx` present, so the `except Exception` path is not what's swallowing it):

```python
# b0-b1 joined, b2-b3 joined, the two pairs never touch; slack at b0
net = pp.create_empty_network()
b = [pp.create_bus(net, vn_kv=110) for _ in range(4)]
pp.create_line_from_parameters(net, b[0], b[1], length_km=1, r_ohm_per_km=.1, x_ohm_per_km=.1, c_nf_per_km=0, max_i_ka=1)
pp.create_line_from_parameters(net, b[2], b[3], length_km=1, r_ohm_per_km=.1, x_ohm_per_km=.1, c_nf_per_km=0, max_i_ka=1)
pp.create_ext_grid(net, b[0])
audit_network(net)
```

```
ISLANDED NETWORK (b2,b3 have no path to the slack at b0):
  status: ok   codes: NONE
ORPHAN BUS (b1 has no lines whatsoever):
  status: ok   codes: NONE

real components        : [{0, 1}, {2, 3}]
union of all components: {0, 1, 2, 3}
in_service buses       : {0, 1, 2, 3}
=> difference flagged  : set()
```

This is the check most likely to save a caller a failed solve — an unsupplied bus is the classic non-convergence cause — so it's worth getting right rather than dropping. What you want is reachability from the slack, not the union of components. pandapower ships this directly:

```python
from pandapower.topology import unsupplied_buses
for idx in sorted(unsupplied_buses(net) & in_service_buses):
    add("error", "DISCONNECTED_BUS", "In-service bus is not supplied by any source.", "bus", int(idx))
```

Either way, please add a test with two islands — the current suite has no disconnected-bus case, which is why this got through CI green.

While you're in there: `except Exception: pass` at `:133` is doing a lot of work silently. If `networkx` is genuinely optional, catch `ImportError` and let real bugs surface; as written it would have hidden this one too, had it raised instead of quietly computing the wrong set.

## Blocking: wrong package

`powermcp/` is the CLI and installer package — `cli.py`, `registry.py`, `runner.py`, `doctor.py`, `wizard.py`. It has no runtime dependency on any solver; it only names them as strings in the registry. Dropping a module there that does `import pandapower.topology` at call time makes the launcher package depend on a solver it currently only launches.

The pandapower bridge is `pandapower/panda_mcp.py`. That's where this belongs.

## Not blocking, but worth settling

- **The feature is unreachable.** Nothing imports `audit_network`, it isn't in `__all__`, and no MCP tool exposes it. The PR description says the point is that "MCP/agent workflows can perform a deterministic validation step before invoking a solver" — but as merged, no agent could call it. A thin `audit_network` tool on the pandapower server would close that, and would also give the checks a real consumer to shape them.
- **The voltage bounds are magic numbers.** `min_vm_pu` valid in `0.0..1.1` and `max_vm_pu` in `0.9..2.0` (`:67-69`) are asserted, not derived. A distribution study with `max_vm_pu=0.88` would be flagged as an error for no stated reason. A comment on where these come from, or looser bounds that only catch the genuinely nonsensical, would age better. I confirmed there are no false positives on `case9`, `case118` or `mv_oberrhein` today, which is good — but that's because those networks leave the columns unset, not because the bounds were validated.
- **`counts["info"]` is always 0** — no check emits `"info"`. Fine to keep the key for shape stability, just noting it's currently decorative.

Happy to look again once the topology check and the module move are sorted.

PR #69, opened 2026-09-16
Test on Python 3.13 and 3.14

samtalki:
## Summary
- Run the test suite on Python 3.13 and 3.14 in addition to 3.10 and 3.12, so the matrix covers the floor of `requires-python` and every version the classifiers claim.
- Add the 3.13 and 3.14 classifiers; the README and the package guide state the tested range.
- Move the checkout and setup-python steps to their current major tags.

Stacked on #67, which stacks on #64; draft until those merge.

## Testing
Full suite from the branch root, no code changes needed on either new interpreter (`powerio` installs from its abi3 wheel):

| Interpreter | Result |
|---|---|
| CPython 3.12.12 | 390 passed, 20 skipped |
| CPython 3.13.14 | 390 passed, 20 skipped |
| CPython 3.14.7 | 390 passed, 20 skipped |


qian-harvard:
Landed on `main` as `19c1e9b`, and CI is green on all four: **py3.10, py3.12, py3.13, py3.14 all success**.

I cherry-picked your commit (`2ce0b77`, with `-x` and your authorship preserved) rather than merging the branch. The reason is that this PR is stacked on #67, so merging it would have pulled the whole Tellegen server in with it, and #67 is a separate call for you and @qian-harvard to make — I've left a note there. Your commit is self-contained (workflow matrix, action versions, two classifiers, README wording), so it lifted out cleanly with no conflicts.

One thing worth checking, which I deliberately did not silently 'fix' for you: the three-way merge of your branch would have kept `main`'s test step, but your commit's own diff was written against a parent where the step was still `pytest tests/ -q`. `c41c525` (which came in with #64) had since widened it to include the PowerFactory suites. The cherry-pick touches only the matrix and the action versions, so that widening is intact on `main` — but if you rebase the rest of the stack, watch that `Run tests` step doesn't get reverted.

Also flagging a small pre-existing inconsistency, since your commit comment now reads "every version the classifiers claim": 3.11 is in the classifiers but not in the matrix. Not introduced here, and I left it alone.

Closing as merged-by-cherry-pick — reopen if you'd rather land it as the branch once #67 resolves.

PR #70, opened 2026-09-16
Unify tool error shapes and remove import-time exits

samtalki:
## Summary
- Add `powermcp/errors.py`: `tool_success`, `tool_error` and `run_tool` build the one result shape every bundled server reports, `{"status": "success", ...}` or `{"status": "error", "message": ...}`. A failure inside a tool reaches the caller as a result, not as an MCP protocol error, and its traceback goes to the log rather than to the stdio channel.
- Move GenX, PowerFactory, PLEXOSDB and OpenDSS onto that shape. GenX and PowerFactory reported `{"success": bool}`, OpenDSS `{"success": bool, "error" | "payload"}`, and PLEXOSDB `{"ok": bool, "error"}` while letting every r2x failure escape as a raw protocol error. Their READMEs and the top-level README state the shape. HOPE keeps its own `{"ok": ...}` results and is not part of this change.
- Load the OpenDSS engine on first use. `core/engine.py` used to build the `DSS()` instance at import, so a missing or broken engine library ended the server before it listed its tools; `get_dss()` now builds and wires it on the first tool call and a failure there is a tool result.
- Keep the LTSpice server importable without its engines. The three module-scope `sys.exit` calls are gone; matplotlib (on the Agg backend) and spicelib are imported inside the tools that need them, and a missing one is reported with its install hint.

Stacked on #67, which stacks on #64; draft until those merge.

## Testing
`python -m pytest tests -q` on CPython 3.12.12 from the branch root: 424 passed, 20 skipped. New tests cover the helper module, the GenX, PowerFactory and PLEXOSDB shapes (with fakes that raise in place of the engines), the OpenDSS server importing and answering with a DSS that refuses to load, and the LTSpice server importing with matplotlib and spicelib absent. Each changed server was imported under the test stubs with stdout captured: nothing is written to the stdio channel at import.

https://claude.ai/code/session_01TbZRpXL3GhRRgXrccNDWka


qian-harvard:
Flagging one concrete blocker ahead of a full review, since it is invisible to CI and will stay that way.

`PLEXOSDB/plexosdb_mcp/main.py:183` now returns `tool_success(...)`, but `PLEXOSDB/tests/test_tools.py:129` still asserts the old shape:

```python
>>> from powermcp.errors import tool_success
>>> result = tool_success(output_path="/tmp/out/system.json", model_name="Base", component_types={"Bus": 2})
>>> result
{'status': 'success', 'output_path': '/tmp/out/system.json', 'model_name': 'Base', 'component_types': {'Bus': 2}}
>>> result["ok"]          # test_tools.py:129 -> self.assertTrue(result["ok"])
KeyError: 'ok'
```

`.github/workflows/test.yml:33` runs `pytest tests/` only, so `PLEXOSDB/tests/` is never collected and this stays green on the dashboard. Same blind spot hides #61's two red tests — there are 67 test functions across `PSCAD/`, `HOPE/`, `PLEXOSDB/`, `PowerFactory/` and `surge/` that CI has never executed. Worth fixing the collection scope in its own PR rather than here.

Two other things for whenever this comes out of draft:

1. **`Agent_DIgSILENT.py:2241`** reads the report key this PR renames, in the standalone entry point. Worth a check that the rename is complete rather than just complete within the MCP layer.
2. **Please redo the PowerFactory conversion on top of #66 rather than before it.** #66 is a semantic change worked out against a live PowerFactory that none of us here has — it adds a `structured=True` passthrough to `_agent_result` so `delete_component` can report `deleted` and `graphics` separately. Your side rewrites the same function mechanically. `merge-tree` puts the two at 15 conflict hunks in `MCP_PowerFactory.py`, and the merged result has to express #66's structured result *in* your error shape, including deciding what `deleted=true, success=false` becomes under `{"status": "error"}`. Reconstructing that from the conflict markers is much harder than redoing the mechanical half on top.

Also: a migration note should name all three dropped keys — `success`, `ok`, and OpenDSS's `error` — since this is a breaking change for any MCP client reading them. And `README.md:292`'s uniformity claim should be scoped to the converted servers, as HOPE deliberately keeps its own shape.

Unrelated heads-up: `main` moved to `v0.3.1` an hour ago (#72, capping `powerio` at `<0.10` — it had been red since 6 Sept). You will pick up trivial one-line conflicts in the 13 pin sites plus `powermcp/__init__.py` and `tests/test_registry.py`; take your side on all of them.

qian-harvard:
#64 is merged (`809be61`), which clears most of this stack. I resolved its fifteen conflicts myself — fourteen were the powerio pin and the version string, and the fifteenth was the hatch `exclude` list, where `main` and the branch had each added different entries and the answer was to keep both. Verified before pushing: 358 passed / 19 skipped on a clean 3.12 venv, plus `build` and `twine check`.

Re-running the merges against the new `main`:

| PR | against new `main` |
|---|---|
| #67 | **clean**, 393 passed / 20 skipped |
| #69 | **clean** |
| #70 | one conflict: `PowerFactory/MCP_PowerFactory.py` |
| #71 | same one conflict |

I'm leaving this one to you, because it isn't a conflict in the mechanical sense — it's a genuine divergence, and resolving it means redoing part of this PR's work rather than picking a side.

This branch predates the whole PowerFactory refactor. `main` has picked up commits the branch never saw:

- `87c5bc9` **centralized application access** — every tool is now `def _impl(app)` with the app passed in. This branch still has `def _impl()` doing `app = DIgSILENTAgent._shared_app` plus its own `None` check, which is the pattern `87c5bc9` removed. Six sites on `main`, nine here.
- `b77eb75` **structured read results** — `_agent_result` grew a `structured: bool = False` parameter. This branch's rewrite drops it entirely in favour of `_reported()`, so the structured path would be lost.
- `f636983` **component-listing optimization** — `main` returns `total_count=len(components)`, `returned_count=len(results)`, `results=results`. This branch returns `total_count=len(results)`, `returned_count=min(len(results), limit)`, `results=results[:limit]`. Those are different numbers with different meanings, and taking the branch side reverts the optimization.
- Plus #66 and #77 on top.

So the merge needs `tool_success`/`tool_error` re-applied *on top of* the refactored server, not the branch's copy dropped over it: keep `main`'s `_impl(app)` signature and its `structured=` path, and decide explicitly how `_reported()` composes with `structured=True` — right now they're two different answers to the same question. The `list_components` counts need a deliberate call too.

I didn't want to guess at that on your behalf: it's your design in #70, the PowerFactory refactor is @aswinkrishnapoyil's, and none of us can exercise the result against a live PowerFactory. Getting it subtly wrong would be invisible until someone runs it.

Rebasing on the new `main` should shrink this to just that one file.

aswinkrishnapoyil:
@qian-harvard I'm working on and I'll get it patched up and push it for review by Monday.

aswinkrishnapoyil:
@qian-harvard @samtalki I’ve completed the rebase and PowerFactory conflict resolution here:
https://github.com/aswinkrishnapoyil/PowerMCP/tree/fix/pr70-review-findings
Head commit: https://github.com/aswinkrishnapoyil/PowerMCP/commit/746e570
The branch is rebased onto current main. The PowerFactory implementation preserves the centralized _impl(app) access, structured result path, optimized component-count semantics, and the changes from #66 and #77.
The follow-up also addresses:
- partial deletion returning external status="error" while retaining deleted and graphics;
- the stale PLEXOSDB ok assertion;
- the standalone PowerFactory report["success"] check;
- documentation of the migrated success, ok, and OpenDSS error fields;
- README scoping for servers not yet converted;
- regression coverage for partial deletion.
Verification completed:
- pytest tests/ -q: 419 passed, 26 skipped
- focused review tests: 17 passed
- live PowerFactory result-shape checks passed, including total_count=39, returned_count=2, and the structured unsupported-category error.
No PowerFactory calculations or network modifications were performed during the live validation.
Since #70’s source branch is samtalki:feat/server-hardening, I have not opened a duplicate PR. @samtalki, would you prefer to update #70 from this branch, or should I open a replacement PR?

qian-harvard:
Heads-up on the rebase burden here, since it grew after my last comment.

#78 (PowerFactory contingency tools) adds 232 lines to `PowerFactory/MCP_PowerFactory.py` — the one file this PR still conflicts on. It takes `structured=True` from one call site to two, and `def _impl(app)` from six to eight. Both are exactly the conventions this PR replaces with `tool_success`/`tool_error`, so when #78 lands there will be three more tools (`run_contingency_analysis`, `list_contingencies`, `get_contingency_results`) to convert on top of the reconciliation I described above.

Nothing wrong with #78 — it's correctly written against current `main`. But if you'd rather not chase a moving file, it may be worth landing the `MCP_PowerFactory.py` reconciliation sooner rather than later, or splitting it out of this PR so the rest of the error-shape work (GenX, PLEXOSDB, OpenDSS, LTSpice) can merge independently of the PowerFactory question. That split would let most of this PR go in now.

Your call — just flagging it rather than letting the conflict quietly get larger.

PR #71, opened 2026-09-16
Adopt the powerio 0.11.3 names

samtalki:
## Summary
- Pin `powerio[...]>=0.11.3,<0.12` everywhere (15 places), and give every test module that needs the 0.11.3 names the same version floor, so an older powerio skips instead of failing.
- Read a module's structural type through `PioModule.type_name` and build diagnostic records with `powerio.diagnostic_records`, both public in powerio 0.11.3. The private attribute read and the two copies in `powermcp/solver_case.py` are gone; `powermcp/tellegen.py` and `pandapower/panda_mcp.py` use the powerio names.

Stacked on #70, which stacks on #67 and #64. Draft until powerio 0.11.3 is on PyPI: until then the pin cannot resolve and CI cannot install the package.

## Testing
`python -m pytest tests -q` on CPython 3.12.12 from the branch root, against a wheel built from the powerio 0.11.3 release stack:

| Environment | Result |
|---|---|
| powerio 0.11.3 stack wheel, all four floored modules running | 424 passed, 20 skipped |
| released powerio 0.11.2 | 302 passed, 18 skipped, the four floored modules skipped with the version message |

https://claude.ai/code/session_01TbZRpXL3GhRRgXrccNDWka


qian-harvard:
powerio 0.11.3 is on PyPI now (`pip index versions powerio` → latest 0.11.3), so the stated reason this was held — "the pin cannot resolve and CI cannot install the package" — no longer applies. CI is green here on py3.10 and py3.12.

The remaining obstacle is the rebase: this conflicts with `main` on the fourteen powerio pin sites (`main` is capped `<0.10` by #72) plus `PowerFactory/MCP_PowerFactory.py` from #66. Details and the full file list in my comment on #64. Given 0.11.3 is out, it may be simpler to land this one directly on a rebased base than to walk #64 → #67 → #70 → #71 in order — your call, since you know what's load-bearing in the middle two.

qian-harvard:
#64 is merged (`809be61`) — I resolved its fifteen conflicts and verified the result (358 passed / 19 skipped, clean 3.12 venv, `build` + `twine check`). powerio now resolves to 0.11.3 under the `>=0.11.2,<0.12` pin that #64 put in, so this PR's 0.11.3 floor is a small step rather than the jump it was.

Against the new `main`, this is down to a single conflict: `PowerFactory/MCP_PowerFactory.py`, inherited from #70. It's a semantic divergence, not a pin — this branch predates `87c5bc9` (centralized `_impl(app)` access), `b77eb75` (the `structured=` result path) and `f636983` (the component-listing counts), so the rewrite here would revert all three. I wrote up the specifics on #70 rather than guess at the resolution.

A rebase on the new `main` should reduce this PR to that one file plus the 0.11.3 name changes.

PR #78, opened 2026-09-19
Add PowerFactory contingency analysis and inspection tools

aswinkrishnapoyil:
## Summary

Adds PowerFactory MCP tools for executing and inspecting the active study case’s configured Contingency Analysis command (`ComSimoutage`).

The new tools are:

- `run_contingency_analysis`
  - Uses the active study case’s existing Contingency Analysis command.
  - Executes the command without replacing or modifying its configured contingencies.
  - Returns the native PowerFactory execution code.
  - Returns the command identity and stable calculation settings.
  - Reports structured errors when no study case or contingency command is available.

- `list_contingencies`
  - Lists configured static fault cases and outage events.
  - Returns each outage’s target element, execution time, action, and disabled state.
  - Does not modify the project.

- `get_contingency_results`
  - Reads the configured AC or DC contingency result file.
  - Returns bounded rows and columns to avoid excessively large MCP responses.
  - Decodes PowerFactory `GetValue()` status/value pairs.
  - Returns unavailable cells as `null` with their native error codes.
  - Always releases the loaded `ElmRes` result file.

All three tools are documented in the PowerFactory tool overview.

## Scope

`run_contingency_analysis` executes an already configured PowerFactory contingency analysis. It deliberately does not create contingency definitions or alter the command configuration.

PowerFactory must have:

- An active project.
- An active study case.
- A `ComSimoutage` command in that study case.
- Valid contingency definitions configured in that command.

`list_contingencies` and `get_contingency_results` are read-only inspection tools. They do not execute calculations or change contingency definitions.

Automatic creation of N-1/N-2 contingencies and modification of `ComSimoutage` settings are intentionally outside the scope of this PR.

## `run_contingency_analysis` result contract

A successful execution returns information in this form:

```json
{
  "success": true,
  "message": "Configured contingency analysis completed",
  "execution_code": 0,
  "command": {
    "name": "Contingency Analysis",
    "class_name": "ComSimoutage",
    "full_name": "..."
  },
  "settings": {
    "data_source": "MAN",
    "calculation_method": 1,
    "linear_method": 0,
    "dynamic_contingencies": 0
  }
}
```

A nonzero native execution code is reported as a structured failure.

## `list_contingencies` result contract

A successful query returns information in this form:

```json
{
  "success": true,
  "total_count": 1,
  "returned_count": 1,
  "results": [
    {
      "name": "MCP N-1 Line Test",
      "class_name": "IntEvt",
      "full_name": "...",
      "outage_count": 1,
      "outages": [
        {
          "name": "Outage Event",
          "class_name": "EvtOutage",
          "target": {
            "name": "Line 01 - 02",
            "class_name": "ElmLne",
            "full_name": "..."
          },
          "time_s": 0.0,
          "action": 0,
          "disabled": false
        }
      ]
    }
  ]
}
```

## `get_contingency_results` result contract

A successful bounded read returns information in this form:

```json
{
  "success": true,
  "calculation_method": "ac",
  "result_file": {
    "name": "Contingency Analysis AC",
    "class_name": "ElmRes",
    "full_name": "..."
  },
  "total_rows": 2,
  "total_columns": 838,
  "returned_rows": 2,
  "returned_columns": 2,
  "truncated": true,
  "columns": [
    {
      "index": 0,
      "variable": "b:i_obj"
    },
    {
      "index": 1,
      "variable": "b:number"
    }
  ],
  "rows": [
    {
      "index": 0,
      "values": [0.0, -1.0]
    },
    {
      "index": 1,
      "values": [0.0, null],
      "errors": [
        {
          "column": 1,
          "code": 3
        }
      ]
    }
  ]
}
```

The requested result slice is limited to 20,000 cells. Larger requests return a structured error.

## Validation

Automated validation:

- All GitHub Actions checks pass on Python 3.10, 3.12, 3.13, and 3.14.
- Focused PowerFactory tests: 31 passed, including 3 subtests.

Live PowerFactory validation:

- Activated `Case 1`.
- Executed its configured `ComSimoutage` command successfully with native execution code `0`.
- Listed one configured static line-outage contingency.
- Verified the outage target as `Line 01 - 02`.
- Read bounded AC results from a result file containing 2 rows and 838 columns.
- Verified scalar decoding and native error-code reporting.
- Verified that the DC result file can be read when empty.
- No calculations were run during the final read-only inspection validation.


aswinkrishnapoyil:
Added the contingency inspection follow-ups in `16bb687`:

- Added `list_contingencies`.
- Added bounded AC/DC `get_contingency_results`.
- Decoded PowerFactory `GetValue()` status/value pairs.
- Represented unavailable cells as `null` with error codes.
- Removed misleading result-column and outage-event paths.
- Added automated coverage.

Validation:
- 31 tests passed, including 3 subtests.
- Live PowerFactory checks passed against `Case 1`.
- AC results: 2 rows × 838 columns, with bounded retrieval verified.
- Static contingency and outage target metadata were verified.
- No calculations were run during the final read-only validation.

qian-harvard (commented):
This is good work, and it's noticeably refactor-aware in a way that made it cheap to review: the three tools use `_impl(app)` and `structured=True` rather than reaching for `_shared_app`, so they sit on `main`'s current conventions instead of fighting them.

Worth saying up front, because it's new: **your PowerFactory tests actually ran in CI this time.** `c41c525` landed with #64 yesterday and widened the test step, so all four interpreters really executed them — `389 passed, 19 skipped, 3 subtests` on 3.10/3.12/3.13/3.14. Before that, PR #77's 29 tests were invisible to CI. I reproduced locally against powerio 0.11.3: 31 passed.

I went after the error paths rather than the happy paths, since those are what a live-PF-free reviewer can actually add. Two held up well:

- An exception raised mid-read (after a successful `Load()`) still hits `finally` and calls `Release()`. No leak.
- An outage with no `p_target` yields `"target": null` rather than blowing up on `_object_summary(None)`.

Three things I'd change, one of which I think is worth fixing before this merges.

## 1. `execution_code` disappears exactly when it matters

On a nonzero native code the whole payload is:

```json
{"success": false, "message": "ComSimoutage returned error code 2"}
```

`execution_code` is gone, and so are `command` and `settings`. The contract section of this PR says the tool "Returns the native PowerFactory execution code", and it does — but only on success. On failure the code survives solely as prose inside `message`, so a model that wants to branch on *which* PF error occurred has to regex an English sentence.

That's the one case where the number is most useful. `raise RuntimeError(...)` at `Agent_DIgSILENT.py:2110` discards the structure; returning the failure dict directly, with `execution_code` and the `command` block intact, keeps it. Small change, and it makes the failure contract match the success contract.

## 2. A configured-but-empty fault case is invisible

`list_contingencies` only collects an `IntEvt` when `outages` is non-empty, and `total_count` is the length of that filtered list. So a project with two fault cases, one of them empty, reports `total_count: 1` and never mentions the other.

Self-consistent, but it means an agent cannot tell "no contingencies are configured" from "a fault case exists and has nothing in it" — and the second is a misconfiguration worth reporting, since it's the likely cause of a contingency run that does nothing. I'd include empty cases with `outage_count: 0`, or add a `skipped_empty` count.

## 3. The `GetValue()` decoding heuristic is undocumented

```python
if isinstance(raw_value, (list, tuple)) and len(raw_value) == 2 and isinstance(raw_value[0], int):
    error_code, value = raw_value
```

Any 2-element sequence whose first element is an `int` is read as a status/value pair. A genuine 2-tuple value of `(1, 7.5)` decodes as `error_code=1` and lands as `null` with a fabricated error entry:

```
values: [42.0, None, 3.3]   errors: [{'column': 1, 'code': 1}]
```

For scalar contingency results I don't think PF ever hands back a 2-tuple *value*, so this is very likely safe in practice — and you verified it live, which counts for more than my reasoning about it. But the safety rests on an assumption the code doesn't state. One comment saying "ElmRes.GetValue returns (status, value); a bare scalar is tolerated for older builds" would stop someone from later extending this to a result type where it isn't true.

## Minor

`MCP_PowerFactory.py:285` has one blank line before `@mcp.tool() def import_project`; the rest of the file uses two.

## One coordination note, not a problem with this PR

This adds 232 lines to `MCP_PowerFactory.py`, which is the single file #70 and #71 still conflict on, and it takes `structured=True` from one call site to two and `_impl(app)` from six to eight. Both are conventions #70 proposes to replace with `tool_success`/`tool_error`. Nothing here is wrong — you're matching current `main`, which is the right thing to do — but landing this first does grow that rebase, so @samtalki should know. I've noted it on #70.

Happy to merge once the `execution_code` bit is sorted; the other two I'm content to leave to your judgement.

qian-harvard:
Pushed the `execution_code` fix to this branch as `a34ccd3` (maintainer edit) so it could go in with the rest rather than round-trip.

A nonzero `ComSimoutage` code now returns the same shape as a success — `execution_code`, `command` and `settings` all intact — instead of collapsing to `{"success": false, "message": ...}`. The message still names the code and the failure is still logged, so nothing is lost on the human-readable side.

I deliberately left the exception path alone: when there's no active study case or no `ComSimoutage`, there's no command object to describe, so the minimal shape is the right answer there. Verified all three paths:

```
Execute()->0: success=True  execution_code=0 command=yes settings=yes
Execute()->2: success=False execution_code=2 command=yes settings=yes
no active study case: {"success": false, "message": "No PowerFactory study case is active"}
```

Added the regression test for the nonzero path; it fails on the previous implementation with `KeyError: 'execution_code'`, which is how I confirmed the gap was real rather than theoretical. CI is green on all four interpreters at **390 passed** (up from 389).

Merging. Thanks for this — the live verification and the scope discipline (executing the configured command rather than inventing contingency definitions) both made it straightforward to review.

The other two points from my review, empty fault cases and the `GetValue()` decoding assumption, are neither blocking nor urgent; I've opened a follow-up issue so they don't get lost.

PR #82, opened 2026-09-21
Install HiGHS and Ipopt so the surge OPF groups run in CI

qian-harvard:
The two solver-dependent surge groups — `Solver matrix` and `Contingency + transfer` — skip on a stock runner. `surge-py` binds the solver C libraries directly, and neither the `highspy` nor the `cyipopt` wheel ships `libhighs.so` or `libipopt.so`.

Skipping is the correct behaviour when the library is genuinely absent (that is what `f4c0ec5` established), but it meant those groups never actually ran anywhere in CI — they were green by abstention.

Both packages are in Ubuntu 24.04, which `ubuntu-latest` currently is; I checked before writing this rather than guessing. Ipopt there is 3.14, inside the 3.x range surge supports.

## What to look for

The signal is the **skip count**, not just a green tick:

| | before | expected after |
|---|---|---|
| py3.12/3.13/3.14 | 445 passed, 21 skipped | 447 passed, 19 skipped |
| py3.10 | 439 passed, 20 skipped | unchanged — `surge-py` cannot install there |

If the count does not move, the libraries are present but surge is not finding or accepting them, and this is not worth keeping.

## Risk

Low. If the libraries are ABI-incompatible with `surge-py`, the probe fails and the groups skip exactly as they do today, so the worst case is the status quo plus ~30s of apt time.

Opened as a PR rather than pushed to main because the workflow only runs on `main` pushes and pull requests, so this is the only way to test it — and I would rather not find out on main.


qian-harvard:
Closing this — I tried it, and it makes things worse rather than better. Writing up what I found so nobody spends the afternoon on it again.

## The premise was wrong

I offered this on the assumption that `apt install libhighs-dev coinor-libipopt-dev` would work. **`libhighs-dev` does not exist in any Ubuntu suite.** My pre-check read HTTP 200 from `packages.ubuntu.com` and I took that as existence; the page was actually "Package not available in this suite", served with a 200. A nonexistent package name returns 200 there too. That one is on me — I checked the status code instead of the content.

Ipopt is genuinely packaged (`coinor-libipopt-dev`, 3.14, inside surge's supported 3.x). But Ipopt alone buys nothing: both skipped groups need an **LP** solver, so they skip on HiGHS regardless.

## No packaged or bundled route exists

The macOS `highspy` wheel ships `libhighs.1.15.dylib` alongside `_core`, which is why these groups pass on a developer Mac — though in my case it was actually Homebrew's `/opt/homebrew/lib/libhighs.dylib` being found. The **Linux** wheel links HiGHS statically into `_core.cpython-3xx-x86_64-linux-gnu.so` and ships no `libhighs.so` at all. surge's error message is accurate on Linux.

That leaves building from source, which is what the second commit here does.

## Building it actively breaks the suite

With `$HOME/highs/lib` on `LD_LIBRARY_PATH`, the built `libhighs.so` shadows what `highspy`'s `_core` was linked against:

```
ImportError: highspy/_core.cpython-312-x86_64-linux-gnu.so:
             undefined symbol: _ZN5Highs13releaseMemoryEv
```

That takes down **`test_pypsa_optimize_investment_uses_modern_optimizer`** and `test_pypsa_optimize_uses_modern_optimizer` — real PowerMCP coverage that passes today, broken in exchange for coverage of somebody else's solver integration.

And it did not even work: skips stayed at **21**, so surge never picked the library up. Result was *2 failed, 443 passed, 21 skipped*, at 3–5 minutes of build per job.

## Recommendation: leave it skipping

Worth being clear about what is actually lost. Those two groups exercise **surge's own solver integration** — DC/AC-OPF, SCOPF, SCED, SCUC across solver selections. The PowerMCP code under test is the MCP tool wrappers, and the other six groups cover those on every run. The marginal coverage here is mostly upstream's concern.

Current behaviour is already correct: a solver that is present but returns a wrong answer still fails the run; only genuine absence skips, with a message saying which library is missing. Developers with Homebrew HiGHS and Ipopt — which is how I found the two silently-failing groups in the first place — get the full run locally.

Not worth 4× a multi-minute build plus a coexistence problem with highspy to buy that back.

